PR #28, opened 2026-03-11
Add UI enhancements: icons, favicon, profile page, clear filters

quevon24:
## Summary
- Add Font Awesome 6 Free icons throughout the UI (header nav, buttons, filter controls)
- Header redesign: favicon logo, icons on nav links, Queue nav link, user dropdown (profile + sign out)
- Add user profile page (/profile/) with first name, last name, email editing
- Add password change page (/profile/password/) using Django's PasswordChangeForm
- Consistent filter layout across scan list, opinion list, and queue: selects on top, buttons right-aligned below
- Use btn-outline style for View links in scan list to match queue and opinions
- Add \`url_replace\` template tag to preserve active filters when switching queue status tabs
- Clear filters in queue resets all params including the active status tab


scanning/assets/static-global/css/font-awesome.min.css
```diff
@@ -0,0 +1,4 @@
+/*!
+ *  Font Awesome 4.7.0 by @davegandy - http://fontawesome.io - @fontawesome
```
grossir: Claude is complaining about this. I double checked and it's correct that this is an old and no longer maintained version of FA. Should this be updated?

>  FA 4.7.0 was released in 2017 and is no longer maintained. The vendored files add ~500KB to the repo (SVG font alone is 2671 lines). Consider:
  - Using Font Awesome 6 Free (actively maintained, better icon set)
  - Loading from a CDN instead of vendoring
  - Using a subset/tree-shaken approach to only include the ~10 icons actually used

scanning/templates/scanning/password_change.html
grossir, line 1: did you consider using Mike's suggestion of a @freelaw email login?

scanning/views.py
```diff
@@ -151,7 +154,7 @@ def scan_detail(request, pk):
     opinion_count = scan.opinions.count()
 
     review_form = None
-    if request.user.is_staff and scan.status != Status.APPROVED:
```
grossir, line 168: dropping the `is_staff` check means any logged in user can perform ("POST") a scan review.

I think you should move that condition here:

        if request.method == "POST" and request_user.is_staff:

grossir (commented):
one important comment on who can POST a ScanReview

scanning/templates/scanning/password_change.html
quevon24: I did consider it, but I haven't discussed it in detail with Bill yet.

scanning/views.py
quevon24: In this case, I removed is_staff because I believe the idea is for them to do everything from the frontend without needing to add permissions or having to go directly to the admin. Now, if we want to use the is_staff flag, give admin access, and have more granular control over actions, we could use is_staff but also create groups with a set of permissions.

For the moment i can remove the profile changes to get this up and running.

What do you think about this?

grossir: @quevon24 

I agree they should do everything on the frontend

The problem is the current code is allowing anyone that can log in, includding the scanners, to review a scan. Meaning, they could approve their own scans?

That's why I was suggestion to add this `if request.method == "POST" and request_user.is_staff`. So, only `is_staff` users can "POST" a review

if it is a "GET", to look at a review, anyone can do it.

quevon24: That is a good question, I would hope that the person who did the scan would be responsible for reviewing it, but I think it is better to ask, for now we can leave it as you mentioned so that only those who have the is_staff can do it, initially they would not have that flag so they cannot see the admin.

quevon24:
I merged main and narrowed the scope to UI changes only. The review form was removed entirely in the process since main no longer has it, so that comment no longer applies.

grossir (approved):
Looks good!

Issue #29, opened 2026-03-12
Add dev data for easier testing on development


grossir:
Currently, the only data loaded when the container is started are the reporters

So, everytime we have to create a superuser or users; and scans, and other relevant objects. This makes it harder to actually explore the webapp in local development. 

I think we should create some placeholder objects so we can more easily test the changes

grossir:
@quevon24 commented this is similar to https://github.com/freelawproject/courtlistener/blob/main/cl/lib/management/commands/make_dev_data.py

PR #30, opened 2026-03-12
feat(make_dev_data): create some dummy data on development environment

grossir:
Solves #29

- Populates the database so local development testing is easier
- creates a staff user (name staff, password password)
- create a scanner user (name scanner, password password)
- create some scans


quevon24 (approved):
Everything looks good, I think it's a good idea to have this command in the entry point to make testing easier.

PR #37, opened 2026-04-02
Add Blackletter

### LINKED ISSUE #36: Merge new interface into current one
@flooie has been working locally on a Django interface using real scanned data and blackletter integration

He will publish the changes as a branch in this repo, so we can adapt it

For this merging to work, we will need to add the [blackletter](https://github.com/freelawproject/blackletter/issues/23) service image into this project docker compose

Some of these issues will be solved by this merge
#32 #33 #34

flooie:
Summary
                                                                                                                 
  This branch adds the full blackletter processing pipeline to the scanning portal — from upload through
  detection, OCR, redaction, opinion splitting, and file generation.                                             
   
  Processing Pipeline                                                                                            
  - Daemon-based queue system (run_daemon management command) for async processing of scans through validate → detect → pair → generate stages                                                                                
  - Integration with blackletter library for detection (CASE_CAPTION, KEY_ICON, CITATION, COURT, DATE, DOCKET,
  PAGE_HEADER), OCR, and PDF generation                                                                          
  - Page management: deletions, insertions, and smart page replacement                                           
  - Margin and redaction rect support for headnote redaction          
  - Opinion metadata extraction from detection bboxes + OCR text (caption, parallel cite, court, date, docket)   
                                                                                                              
  Frontend                                                                                                       
  - 3-step scan processing UI (scan_process.html) with PDF viewer, detection overlays, opinion pairing sidebar,  
  and file generation controls                                                                                   
  - Opinion list and detail pages with pdf.js viewer                                                             
  - Volume queue pages for tracking scanning assignments                                                         
  - Keyboard navigation, detection approval/deletion, and interactive redaction editing                          
                                                                                                                 
  Storage Architecture                                                                                           
  - Split storage backends: PrivateS3Storage for archival of original uploads, LocalProcessingStorage for all    
  pipeline working files                                                                                         
  - Ensures .path works throughout the daemon pipeline regardless of deployment environment
  - Processing files (bitonal, OCR, detections, opinions) always stay local; only originals go to S3             
                                                                                                                 
  Models                                                                                                         
  - New models: Detection, Issue, Volume, PageDeletion, PageInsert, LLMScan                                      
  - Extended Scan with pipeline state fields (stage, progress, output_dir, opinions_json, margin/redaction rects,
   page map, OCR results)                                                                                        
  - Extended OpinionScan with metadata fields and opinion ordering                                               
  - Squashed migrations 0002-0011 into single 0002 with .sql file


docker/scanning/docker-compose.yml
```diff
@@ -51,6 +74,7 @@ services:
     user: root
     volumes:
       - ${SCANNING_BASE_DIR:-../../}:/opt/scanning
+#      - ${BLACKLETTER_DIR:-../../../blackletter}:/opt/blackletter
```
grossir: why is this commented out?

scanning/management/commands/import_scanlist.py
```diff
@@ -0,0 +1,196 @@
...
+                reporter = reporters.get(slug)
+                if not reporter:
+                    self.stderr.write(f"Unknown reporter slug: {slug}")
+                    errors += 1
+                    continue
+
+                # Parse volume number (strip trailing A/B/C)
+                vol_digits = "".join(c for c in volume_str if c.isdigit())
+                if not vol_digits:
+                    self.stderr.write(f"Bad volume: {volume_str}")
```
grossir: we should use `logger` for better integration with sentry

scanning/management/commands/import_scanlist.py
```diff
@@ -0,0 +1,196 @@
...
+                    volume=vol_num,
+                    part_label=part_label,
+                    source=source,
+                    start_page=start,
+                    end_page=end,
+                    notes=combined_notes,
+                    source_library=source_library,
+                    status=Status.UPLOADED
+                    if queue_status == QueueStatus.COMPLETE
+                    else Status.UPLOADED,
```
grossir: the conditional has no effect?

scanning/services.py
```diff
@@ -0,0 +1,2194 @@
...
+
+Functions in this module run outside the request/response cycle, in the
+daemon process.  They must NOT import Django HTTP machinery
+(HttpResponse, render, redirect, etc.).
+"""
+
+import json
+import os
+import re
+import re as re_mod
```
grossir: clean up duplicated import

scanning/services.py
grossir, line 1: this file can probably be broken up

Functions like `_has_s3_credentials` can go into a utils file

scanning/urls.py
```diff
@@ -5,25 +5,199 @@
...
+    approve_scan,
+    assign_page,
+    bake_redactions,
+    cancel_processing,
+    claim_scan,
+    compute_redactions_api,
+    delete_detection,
+    delete_page,
+    dismiss_issue,
+    # dismiss_issues,
```
grossir: should delete commented line?

scanning/views.py
```diff
@@ -307,3 +304,1758 @@ def opinion_upload(request):
...
+    if last_page:
+        scan.end_page = int(last_page)
+
+    scan.has_state_abbrev = "has_state_abbrev" in request.POST
+    scan.uploaded_by = request.user
+    scan.status = Status.UPLOADED
+    scan.save()  # Save first to get a PK
+
+    # Create processing directory and save original PDF there
+    from pathlib import Path as _P
```
grossir: i don't think this should be done. `Path` is a short name, this is just obscuring the import

scanning/models.py
```diff
@@ -210,7 +388,75 @@ class Scan(AbstractDateTimeModel):
...
+        help_text="JSON: per-page redaction rects in image pixels.",
+    )
+    page_count = models.PositiveIntegerField(default=0)
+    has_issues = models.BooleanField(default=False)
+    checked = models.BooleanField(default=False)
+    redacted_pdf_path = models.CharField(
+        max_length=1024, blank=True, default=""
+    )
+    has_state_abbrev = models.BooleanField(default=True)
+    source_library = models.CharField(max_length=200, blank=True, default="")
```
grossir, line 538: these are missing help text?

scanning/models.py
```diff
@@ -348,3 +618,184 @@ def __str__(self):
...
+    """
+
+    scan = models.ForeignKey(
+        Scan,
+        on_delete=models.CASCADE,
+        related_name="detections",
+    )
+    page_index = models.PositiveIntegerField(db_index=True)
+    label = models.CharField(max_length=50)
+    label_id = models.SmallIntegerField()
```
grossir, line 846: available labels should be stored somewhere?

scanning/models.py
```diff
@@ -348,3 +618,184 @@ def __str__(self):
...
+
+    class Meta:
+        unique_together = ("scan", "pdf_page")
+
+    def __str__(self):
+        return f"Delete PDF p.{self.pdf_page} from {self.scan}"
+
+
+class LLMScan(AbstractDateTimeModel):
+    """A masked PDF batch sent to an LLM for text extraction."""
```
grossir: shouldn't this be on CL side?

docker/scanning/docker-compose.yml
```diff
@@ -34,9 +34,32 @@ services:
...
+  #     - ${SCANNING_BASE_DIR:-../../}:/opt/scanning
+  #   networks:
+  #     - scanning_net
+  #   env_file:
+  #     - ../../.env.dev
+
   # Django dev server
   scanning-django:
     container_name: scanning-django
+    platform: linux/amd64
```
grossir, line 68: we don't do this on CL. Why is it needed?

scanning/views.py
```diff
@@ -307,3 +304,1758 @@ def opinion_upload(request):
...
+                x1=det["bbox"][2],
+                y1=det["bbox"][3],
+                img_width=det.get("img_width", 0),
+                img_height=det.get("img_height", 0),
+                model_name="manual",
+                model_count=1,
+                found_by=json.dumps([{"model": "manual", "confidence": 1.0}]),
+            )
+        except Exception:
+            pass
```
grossir: why would this fail? 
If failure modes are unknown, we need error logging, not silent skipping

scanning/views.py
```diff
@@ -307,3 +304,1758 @@ def opinion_upload(request):
...
+            )
+        ]
+        det_path.write_text(json.dumps(existing))
+    return JsonResponse({"status": "ok", "deleted": count})
+
+
+@login_required
+@require_POST
+def add_single_detection(request, pk):
+    """Add or boost a single detection for a scan.
```
grossir: we need documentation on "boosting"

scanning/views.py
```diff
@@ -307,3 +304,1758 @@ def opinion_upload(request):
         "scanning/opinion_upload.html",
         {"form": form},
     )
+
+
+# ---------------------------------------------------------------------------
+# Queue views
+# ---------------------------------------------------------------------------
```
grossir, line 297: we can probably split apart the different types of "views", to make this file shorter + more readable

scanning/views.py
```diff
@@ -307,3 +304,1758 @@ def opinion_upload(request):
...
+    :return: File response streaming the PDF.
+    """
+    scan = get_object_or_404(Scan, pk=pk)
+    if scan.output_dir:
+        ocr = find_ocr_pdf(scan.output_dir)
+        if ocr:
+            return FileResponse(
+                open(ocr, "rb"), content_type="application/pdf"
+            )
+        from pathlib import Path as _P
```
grossir: again, let's just perform a search and normalzie this

grossir (commented):
This is a very big PR. Some comments below from reading plain code, I still need to check it out on the docker environment

We should probably take advantage of the reviewing process to 
- write actual documentation
- refactor code so it's readable / findable

quevon24:
These are all the changes I implemented:

- Moved ensure_output_dir and has_s3_credentials from services.py to utils.py for reuse
- Added pyrightconfig.json for venv-aware type checking (LSP support)
- Removed duplicate re import alias in services.py
- Replaced inline Path as _P imports with top-level import in views.py
- Split views.py into views_process.py (process viewer and actions) and views_api.py (JSON endpoints and file serving)
- Grouped URL patterns by view file
- Added missing badge-queued CSS class and safelisted all dynamic badge classes in Tailwind config
- Changed pipeline completion status from APPROVED to PENDING_REVIEW so users must explicitly approve
- Reset s3_uploaded flag when files are modified (reprocessing, re-pairing, regenerating, etc.) to ensure stale files get re-uploaded  
- Skip S3 HEAD checks on re-upload after changes (only skip existing on retry of partial uploads)
- Added S3 upload warning alert on step 3 when files haven't been uploaded
- Added tooltips to all action buttons on the process page
- Added "Admin" link for staff users on the queue detail page
- Updated tests to match new upload and status behavior
- Update docker-compose.yml file to handle dev credentials for S3, See: https://github.com/freelawproject/courtlistener/pull/5552
- Enable daemon service

grossir (commented):
Sorry, Claude auto posted its review. I will check its comments, triage them, add my own

scanning/models.py
```diff
@@ -210,7 +388,75 @@ class Scan(AbstractDateTimeModel):
         related_name="scans",
     )
     processed_at = models.DateTimeField(null=True, blank=True)
+    queued_action = models.CharField(
+        max_length=30,
+        blank=True,
+        default="",
+        help_text="Action for daemon to run: full_pipeline, detect, reprocess, generate_files, validate.",
+    )
```
grossir, line 488: these should be controlled via an enumeration, as usually done. The enumeration can be also used to explain what each one is about

scanning/views_api.py
```diff
@@ -0,0 +1,864 @@
...
+    if scan.redacted_pdf_path and os.path.isfile(scan.redacted_pdf_path):
+        return FileResponse(
+            open(scan.redacted_pdf_path, "rb"), content_type="application/pdf"
+        )
+    if scan.output_dir:
+        for f in sorted(Path(scan.output_dir).glob("*.pdf")):
+            if "redacted" not in f.name and "bitonal" not in f.name:
+                return FileResponse(
+                    open(f, "rb"), content_type="application/pdf"
+                )
```
grossir: I find this very strange. If we are looking for the redacted PDF. 

But if it is not found, we return any other file, and we precisely exclude the pdfs with "redacted" in name?

scanning/services.py
```diff
@@ -0,0 +1,2135 @@
...
+
+    bucket = settings.AWS_PRIVATE_STORAGE_BUCKET_NAME
+    s3_client = boto3.client("s3")
+    for local_path, s3_key in files_to_upload:
+        s3_client.upload_file(str(local_path), bucket, s3_key)
+
+    Scan.objects.filter(pk=scan_pk).update(
+        s3_uploaded=True,
+        s3_path=s3_prefix,
+    )
```
grossir, line 2141: once this happens, everything on the "output_dir" should be deleted. Otherwise, we are risking having no space left on disk

scanning/services.py
```diff
@@ -0,0 +1,2135 @@
...
+    # Verify final files exist
+    redacted_dir = output_dir / "redacted"
+    masked_dir = output_dir / "masked"
+    if not redacted_dir.is_dir() or not list(redacted_dir.glob("*.pdf")):
+        return "Before approving you need to generate the files."
+
+    # Build S3 prefix
+    short = scan.reporter.short_name
+    start = scan.start_page or 1
+    s3_prefix = f"approved/{short}/{scan.volume}/{start}/"
```
grossir: the whole s3_prefix can be computed from `Scan` metadata. Basically

f"approved/{ scan.reporter.short_name}/{scan.volume}/{scan.start_page or 1}/"

So, why persist this in the DB? It should be a computed property. Currently it is a model field

scanning/utils.py
```diff
@@ -0,0 +1,117 @@
...
+        output_dir = Path(settings.MEDIA_ROOT) / "processed" / str(scan.pk)
+        if scan.reporter and scan.volume:
+            output_dir = (
+                output_dir
+                / scan.reporter.short_name
+                / str(scan.volume)
+                / str(scan.start_page or 1)
+            )
+        output_dir.mkdir(parents=True, exist_ok=True)
+        Scan.objects.filter(pk=scan.pk).update(output_dir=str(output_dir))
```
grossir: Again, Scan.output_dir should be a computed property, not a model field

scanning/utils.py
```diff
@@ -0,0 +1,117 @@
...
+
+    Uses scan.output_dir if set, otherwise falls back to
+    MEDIA_ROOT/processed/<pk>.
+
+    :param scan: The Scan instance.
+    :return: The resolved output directory.
+    """
+    if scan.output_dir:
+        return Path(scan.output_dir)
+    return Path(settings.MEDIA_ROOT) / "processed" / str(scan.pk)
```
grossir: this is redundant, could be included in `ensure_output_dir`

scanning/models.py
```diff
@@ -210,7 +388,75 @@ class Scan(AbstractDateTimeModel):
...
+        default="",
+        help_text="JSON: per-page margin rects in PDF points.",
+    )
+    redaction_rects = models.TextField(
+        blank=True,
+        default="",
+        help_text="JSON: per-page redaction rects in image pixels.",
+    )
+    page_count = models.PositiveIntegerField(default=0)
+    has_issues = models.BooleanField(default=False)
```
grossir: this seems to never be used, can be dropped (searched for .has_issues accross the branch)

Also, accesible via related Issue rows

scanning/models.py
```diff
@@ -210,7 +388,75 @@ class Scan(AbstractDateTimeModel):
...
+    checked = models.BooleanField(default=False)
```
grossir: this is set to True once, but never accessed again. Is it necessary?

scanning/models.py
```diff
@@ -348,3 +618,184 @@ def __str__(self):
...
+    label = models.CharField(max_length=50)
+    label_id = models.SmallIntegerField()
+    confidence = models.FloatField()
+    x0 = models.FloatField()
+    y0 = models.FloatField()
+    x1 = models.FloatField()
+    y1 = models.FloatField()
+    img_width = models.PositiveIntegerField(default=0)
+    img_height = models.PositiveIntegerField(default=0)
+    model_name = models.CharField(max_length=20, blank=True, default="")
```
grossir: this should be a controlled value too

scanning/models.py
```diff
@@ -348,3 +618,184 @@ def __str__(self):
...
+        WARNING = "warning", "Warning"
+        INFO = "info", "Info"
+
+    scan = models.ForeignKey(
+        Scan,
+        on_delete=models.CASCADE,
+        related_name="issues",
+    )
+    page_number = models.PositiveIntegerField(null=True, blank=True)
+    check_name = models.CharField(max_length=100)
```
grossir: this should be a controlled value. Seen:  auto_corrected, suppress_detection, add_detection, approve_detection, process_flag, duplicate_page, no_page_number

Each should be explained

grossir (commented):
Some model changes suggestions. Should help declutter the PR

scanning/models.py
```diff
@@ -348,3 +618,184 @@ def __str__(self):
...
+    y1 = models.FloatField()
+    img_width = models.PositiveIntegerField(default=0)
+    img_height = models.PositiveIntegerField(default=0)
+    model_name = models.CharField(max_length=20, blank=True, default="")
+    model_count = models.PositiveSmallIntegerField(default=1)
+    found_by = models.TextField(
+        blank=True,
+        default="",
+        help_text="JSON: per-model confidence breakdown.",
+    )
```
grossir: Postgres support JSONB fields. This should be one. It would save all the json.dumps / json.loads cycles around the code

```

  ┌────────────────────────────┬───────────────────────────────────────────┐
  │           Field            │                 Reasoning                 │
  ├────────────────────────────┼───────────────────────────────────────────┤
  │ Detection.found_by         │ Stores JSON array, parsed with json.loads │
  ├────────────────────────────┼───────────────────────────────────────────┤
  │ Issue.metadata             │ Stores JSON, parsed with json.loads       │
  ├────────────────────────────┼───────────────────────────────────────────┤
  │ LLMScan.extracted_metadata │ Stores JSON, parsed with json.loads       │
  ├────────────────────────────┼───────────────────────────────────────────┤
  │ Scan.ocr_results           │ Stores JSON, parsed with json.loads       │
  ├────────────────────────────┼───────────────────────────────────────────┤
  │ Scan.opinions_json         │ Stores JSON, parsed with json.loads       │
  ├────────────────────────────┼───────────────────────────────────────────┤
  │ Scan.margin_rects          │ Stores JSON, parsed with json.loads       │
  ├────────────────────────────┼───────────────────────────────────────────┤
  │ Scan.redaction_rects       │ Stores JSON, parsed with json.loads       │
  ├────────────────────────────┼───────────────────────────────────────────┤
  │ Scan.page_map              │ Stores JSON, parsed with json.loads       │
  └────────────────────────────┴───────────────────────────────────────────┘


```

scanning/models.py
```diff
@@ -158,17 +309,41 @@ def __call__(self, instance, filename):
...
+        blank=True,
+        related_name="scans",
+        help_text="The Volume this scan belongs to.",
+    )
     reporter = models.ForeignKey(
         Reporter,
         on_delete=models.PROTECT,
         related_name="scans",
     )
     volume = models.PositiveIntegerField(validators=[MinValueValidator(1)])
```
grossir, line 416: Scan.volume_obj already contains the "reporter" and "volume" information. So these fields are just for redundant?

scanning/models.py
```diff
@@ -158,17 +309,41 @@ def __call__(self, instance, filename):
 
 
 class Scan(AbstractDateTimeModel):
+    volume_obj = models.ForeignKey(
+        Volume,
+        on_delete=models.CASCADE,
```
grossir: `on_delete=models.PROTECT` would be best. We are risking deleting all associated Scans if we delete a volume object

scanning/templates/scanning/scan_process.html
```diff
@@ -0,0 +1,808 @@
...
+function _highlightSidebarOpinion(idx) {
+    var cards = document.querySelectorAll('.opinion-nav');
+    cards.forEach(function(c, i) {
+        if (i === idx) {
+            c.classList.add('selected');
+            c.scrollIntoView({behavior: 'smooth', block: 'nearest'});
+        } else {
+            c.classList.remove('selected');
+        }
+    });
```
grossir: this script should be in a separate file

scanning/static/scanning/viewer.js
```diff
@@ -0,0 +1,735 @@
...
+            if (confirm('Delete PDF page ' + pdfPage + '?')) {
+                deletePage(pdfPage, div);
+            }
+        });
+
+        return div;
+    }
+
+    // --- Delete page ---
+    function deletePage(pdfPage, pageDiv) {
```
grossir: this is duplicated in process_viewer.js

scanning/static/scanning/viewer.js
```diff
@@ -0,0 +1,735 @@
...
+
+            var ctx = overlay.getContext('2d');
+            ctx.clearRect(0, 0, overlay.width, overlay.height);
+            drawExistingRedactions(overlay, redactions[String(pageNumber)], SCALE);
+            rebuildRedactionDivs(pageDiv, pageNumber);
+        });
+    }
+
+    function drawExistingRedactions(overlay, pageRedactions, scale) {
+        var ctx = overlay.getContext('2d');
```
grossir: In fact, all of these are duplicated on process_viewer.js

deletePage(), drawExistingRedactions(), rebuildRedactionDivs(), toggleRedactionMode(), dismissIssue(), deleteDuplicates()

scanning/templates/scanning/scan_process.html
```diff
@@ -0,0 +1,808 @@
...
+            method: 'POST',
+            headers: {'Content-Type': 'application/json', 'X-CSRFToken': csrfToken},
+            body: JSON.stringify({pdf_page: p}),
+        }).then(function(r) { return r.json(); });
+    })).then(function() {
+        btn.textContent = 'Deleted';
+        btn.disabled = true;
+        btn.closest('.issue-card').style.opacity = '0.3';
+    });
+}
```
grossir: and maybe this file is holding duplicates of the other JS files too

scanning/static/scanning/viewer_step2.js
```diff
@@ -0,0 +1,2157 @@
...
+            _pageIndexToLogical[entry.pdf_index] = entry.logical_number;
+            _logicalToPageIndex[entry.logical_number] = entry.pdf_index;
+        }
+    });
+    function _pageNumForIndex(pageIndex) {
+        return _pageIndexToLogical[pageIndex] || (pageIndex + 1);
+    }
+    function _pageIndexForNum(pageNum) {
+        return (_logicalToPageIndex[pageNum] !== undefined) ? _logicalToPageIndex[pageNum] : (_pageIndexForNum(pageNum));
+    }
```
grossir, line 33: this is infinite recursion? 

If `_logicalToPageIndex[pageNum]` is undefined it calls itself with exactly the same input argument?

docker/scanning/docker-compose.yml
quevon24: We used that bind mount to test local blackletter changes during development (installing it as editable inside the container). It's commented out because the Docker image now installs blackletter from PyPI via uv, so the local mount is no longer needed. I can remove it if you prefer.

docker/scanning/docker-compose.yml
quevon24: The `platform: linux/amd64` line forces the container to run as x86 even on ARM Macs. Without it, PaddlePaddle and some PyTorch wheels don't have ARM Linux builds and the container fails to start

scanning/services.py
quevon24: Good point. After a successful S3 upload we can delete the generated output files (redacted/, masked/, unredacted/, and the full redacted PDF) since those are now on S3 and can be regenerated if needed. The processing files (bitonal.pdf, OCR PDF, detections.json, stamped.pdf, images/, original PDF) should stay so we can use the viewer to fix detections or redactions without re-running the full pipeline

grossir: Good point about the files being needed for the viewer. I am not sure what paths the files are being saved on, but if you want to persist them you have to put them on the persistent storage volume. Otherwise, when a PR merge updates the pod image, they will be wiped out @quevon24

scanning/static/scanning/viewer.js
quevon24: Looked into the duplicated JS functions. Here's the breakdown:

Already moved to shared.js: deletePage() and drawExistingRedactions() (these were truly identical).

Same name but not actually duplicates:

- rebuildRedactionDivs() - viewer.js uses /save-redaction-rect/ while process_viewer.js uses /redaction/{id}/delete/. Different endpoints and payload formats.
- toggleRedactionMode() - process_viewer.js has coordinate conversion (px to pt), opinion-edit mode for step 3, and sends the correct {page_index, action, adjusted} format. viewer.js sent        
{page_number, x, y, width, height} which was broken. I removed the redact/whiteout buttons from step 1 since they were non-functional there (no detections or redactions at that stage), so the    
viewer.js version is now dead code.

quevon24:
These are all the changes I implemented:

- Used logger instead of stderr in import_scanlist, removed dead conditional
- Added QueuedAction enum for scan processing actions
- Added CheckName enum for validation issue types
- Added ModelName enum for detection model sources
- Converted 7 JSON TextFields to native JSONField (removed ~60 json.dumps/loads calls)
- Removed unused checked and has_issues fields from Scan model
- Changed volume_obj foreign key from CASCADE to PROTECT
- Made LocalProcessingStorage read MEDIA_ROOT lazily (fixes test file leaks)
- Made output_dir a computed property (removed DB field, simplified utils)
- Updated all views to check directory existence instead of field truthiness
- Documented detection workflow and boosting in CLAUDE.md
- Replaced silent exception with logger.exception for Sentry integration
- Clarified serve_redacted_pdf fallback logic
- Cleaned up generated files after S3 upload (prod only)
- Extracted shared JS functions (deletePage, drawExistingRedactions) into shared.js
- Removed broken redact/whiteout buttons from step 1 viewer
- Added undo for page deletions with confirmation dialog
- Renamed JS files: viewer.js → viewer_step1.js, process_viewer.js → viewer_step2.js
- Fixed infinite recursion in _pageIndexForNum
- Fixed ruff lint errors (import order, unused variables)

There are still a few changes pending

scanning/models.py
quevon24: You're right they're redundant, but scan.reporter and scan.volume are used in ~40 places and volume_obj is nullable. Removing them means null checks everywhere. If we make volume_obj required first, it would be easier to replace the fields with properties that delegate to it. What do you think?

scanning/services.py
quevon24: I didn't know the pod image update wipes those files. Right now everything lives under MEDIA_ROOT/processed/{pk}/... inside the container. If that's not on a persistent volume, we'd lose the processing files on every deploy, and we wouldn't be able to go back and fix detections, adjust redactions, or regenerate files without re-running the full pipeline from scratch.             

We should mount a persistent volume for the processed directory so those files survive deploys. Wiping everything and only keeping S3 uploads would be cheaper on storage but means losing the ability to make corrections without reprocessing.

What do you think is the best option? To use persistent volume we need to tweak something in the code or it is just a mount pointing to a path inside the container?

scanning/models.py
quevon24: Yes, the LLM extraction should be on the CL side. The LLMScan model here just creates PENDING records during file generation but nothing actually calls any LLM API. We're planning to do the extraction in CourtListener so we should remove LLMScan from scanning and handle it entirely on CL's side once we're ready. For now I can remove the model and the record creation in run_generate_files to keep things clean.

scanning/services.py
mlissner: It's easy to mount a drive, but why not use S3 if we want to keep these files around? It's much cheaper?

quevon24:
I gave it a new review and these are all the changes I implemented, I also forgot to push some changes I did on friday:

- Add keyboard shortcuts documentation for the process viewer
- Read detection data from data- attributes instead of inline JS params
- Use data-page attribute for goToPage calls
- Remove LLMScan model  
- Fix opinions_json="" to [] in cancel_processing (JSONField type mismatch)
- Wrap all json.loads(request.body) calls with try/except, return 400 on malformed JSON
- Move SecurityMiddleware to first position in MIDDLEWARE
- Fix daemon scan pickup race with select_for_update(skip_locked=True)
- Fix claim_scan race with conditional UPDATE
- Use Path.open("rb") in PDF-serving views for consistency
- Guard int()/float() on query params to return 400 instead of 500
- Fix dead code re-reading step query param (default to 1)
- Use PENDING_REVIEW instead of APPROVED when cancelling PROCESS stage
- Use .isdigit() guard on page range inputs in queue_upload
- Replace hardcoded /scans/<pk>/process/ redirects with reverse()
- Validate content_type in add_page_insert to reject non-image files
- Extract _caption_is_continuation to module level
- Remove LLMScan references from companion SQL migration

scanning/services.py
```diff
@@ -0,0 +1,2128 @@
...
+    :param scan_pk: Primary key of the scan to edit.
+    :param logical_page_number: The logical page number to insert at.
+    :param image_path: Path to the image or PDF file to insert.
+    """
+    scan = Scan.objects.get(pk=scan_pk)
+    output_dir = Path(scan.output_dir) if scan.output_dir else None
+
+    # Determine insertion position from page_map
+    page_map = scan.page_map
+    insert_idx = len(fitz.open(scan.pdf_path))  # default: append
```
grossir: we should ensure the file is closed after opening it

with fitz.open(scan.pdf_path) as f:
      insert_idx = len(f)

scanning/views_api.py
```diff
@@ -0,0 +1,889 @@
...
+    ocr_pdf = find_ocr_pdf(output_base) if output_base.is_dir() else None
+    if not ocr_pdf:
+        return JsonResponse([], safe=False)
+    from blackletter.margins import compute_margin_rects
+
+    from scanning.services import _adjust_margins_for_detections
+
+    rects = compute_margin_rects(ocr_pdf)
+    rects = _adjust_margins_for_detections(rects, output_base)
+    Scan.objects.filter(pk=pk).update(margin_rects=rects)
```
grossir, line 136: `Scan` has date_created and date_modified that use the auto_now arguments. These work through [signals](https://docs.djangoproject.com/en/6.0/ref/models/fields/#django.db.models.DateField.auto_now) so using .update will break them 

Should change to .save()

It won't affect performance, anyway, since we already have the target scan in memory, it is still a single query

scanning/views_api.py
```diff
@@ -0,0 +1,889 @@
...
+                        "y0": round(adjusted["y0"], 1),
+                        "x1": round(adjusted["x1"], 1),
+                        "y1": round(adjusted["y1"], 1),
+                        "fill": fill,
+                        "type": rect_type,
+                    }
+                )
+                found = True
+                break
+    Scan.objects.filter(pk=pk).update(redaction_rects=rects)
```
grossir, line 220: same as above comment. In fact, there are many instances of this happening

at least 6 searching for "Scan.objects.filter(pk=pk).update("

scanning/static/scanning/viewer_step2.js
```diff
@@ -0,0 +1,2125 @@
...
+        var ps = parseInt(parts[0]), pe = parseInt(parts[1] || parts[0]);
+        if (isNaN(ps) || isNaN(pe) || ps < 1 || pe < ps) {
+            alert('Invalid range');
+            return;
+        }
+        var modelSelect = document.getElementById('model-select');
+        var model = modelSelect ? modelSelect.value : 'medium';
+        btn.textContent = 'Processing...';
+        btn.disabled = true;
+        fetch('/scans/' + documentId + '/reprocess-section/', {
```
grossir: `/reprocess-section/` doesnt exist?

This one does:

```
    path("scans/<int:pk>/reprocess/", reprocess, name="reprocess"),

```

scanning/static/scanning/viewer_step2.js
```diff
@@ -0,0 +1,2125 @@
...
+            allDetections.push(detData);
+            _cancelDetDraw();
+            detectionsVisible[pageNum] = true;
+            pageDiv.querySelector('.detect-btn').classList.add('active');
+            drawDetectionOverlay(pageDiv, pageNum);
+
+            // Auto re-pair if pairing label
+            var _pLabels = ['CASE_CAPTION', 'KEY_ICON'];
+            if (_pLabels.indexOf(labelName) >= 0) {
+                fetch('/scans/' + documentId + '/repare-opinions/', {
```
grossir: this endpoint doesn't seem to exist

scanning/static/scanning/viewer_step2.js
```diff
@@ -0,0 +1,2125 @@
...
+                    var nW = parseFloat(div.style.width);
+                    var nH = parseFloat(div.style.height);
+                    var newBbox = [
+                        Math.round(nL / scaleX),
+                        Math.round(nT / scaleY),
+                        Math.round((nL + nW) / scaleX),
+                        Math.round((nT + nH) / scaleY),
+                    ];
+                    // Update detection in detections.json
+                    fetch('/scans/' + documentId + '/update-detection-bbox/', {
```
grossir: this endpoint doesn't seem to exist

scanning/views_process.py
```diff
@@ -0,0 +1,719 @@
...
+    except json.JSONDecodeError:
+        return JsonResponse({"error": "Invalid JSON"}, status=400)
+    issue_id = data.get("issue_id")
+    Issue.objects.filter(pk=issue_id, scan=scan).delete()
+    return JsonResponse({"status": "ok"})
+
+
+@login_required
+@require_POST
+def dismiss_issues(request, pk):
```
grossir: no URL points to this view

grossir (commented):
@quevon24  Some comments below

Also, important to address Mike's comment on saving intermediate files on S3

scanning/views_api.py
quevon24: I added a custom queryset on AbstractDateTimeModel that stamps auto_now fields on update(), and also overrides save() to include auto_now fields when `update_fields` is passed.

Went structural instead of converting each call site to .save() because it depends on context: in views we already have the Scan in memory (fine to .save()), but in services.py many functions  only receive scan_pk and use .update() to avoid an extra SELECT. Converting those would double the query count. The queryset/save override fixes the auto_now bug for every AbstractDateTimeModel subclass in one place, without forcing a choice between correctness and query count.

quevon24:
About Mike's comment on saving intermediate files on S3, I thought of it this way:

Push the processing files (bitonal, OCR PDF, detections.json, stamped, images, original PDF) to S3 after processing and download them to a local temp directory on demand when the viewer reopens a scan, so pod deploys don't wipe them and we avoid attaching a persistent volume.

@grossir @mlissner what do you think about this?

mlissner:
Why download them to a temp directory? S3 can just host them?

quevon24:
> Why download them to a temp directory? S3 can just host them?

Let me explain why we need to keep these intermediate files around in the first place, then the local vs S3 part makes more sense.

The pdf viewer isn't just a read-only display, users actively edit the scan after processing finishes:

- Drawing/deleting/approving detection boxes updates detections.json
- Marking new redactions (e.g. a user spots something the model missed) needs the OCR PDF + detections to reapply redactions and regenerate the stamped/redacted outputs
- Deleting pages needs the PDF + page images + detections to stay in sync
- Opinion pairing adjustments read detections.json and regenerate paired outputs

If we only kept the final outputs, any fix like a missed redaction spotted days later would force us to re-run the entire pipeline from the original scan (OCR, detection, pairing, etc.), which is expensive. Keeping the intermediates (bitonal, OCR PDF, detections.json, page images, stamped) lets us patch just what changed.

Now, on local vs S3: most of these edits use PyMuPDF (fitz.open()), which works on a filesystem path or a full in-memory buffer. A single user action often touches the PDF + detections.json +  images together, and edits are interactive (many small writes). Pulling the files once to a temp dir per viewer session is much faster than round-tripping to S3 on every click. S3 is the source of truth, local disk is just a short-lived cache for the active editing session.

mlissner:
> Now, on local vs S3: most of these edits use PyMuPDF (fitz.open()), which works on a filesystem path or a full in-memory buffer. A single user action often touches the PDF + detections.json + images together, and edits are interactive (many small writes). Pulling the files once to a temp dir per viewer session is much faster than round-tripping to S3 on every click. S3 is the source of truth, local disk is just a short-lived cache for the active editing session.

Hm, sorry to wade into this, but is the idea that:

1. The user loads a page, which shows the PDF with redactions and whatnot. 
2. The user can make changes via that page (remove pages from the PDF, tweak redactions, etc)
3. Those changes are submitted to the backend as a post.
4. The PDF is then updated with the changes.

And so is the idea that you load the PDF from S3 in step 1, then serve it to the user, and then in step 4 you still have it around on disk to apply the changes?

quevon24:
With the laatest changes, this is wow this PR handles processing files, by env:

**DEVELOPMENT=True (local dev):**
- Everything lives under `MEDIA_ROOT/processed/{pk}/{reporter}/{vol}/{start}/`.
- No S3 calls. No cleanup. Files persist across container rebuilds via the project's bind-mounted MEDIA_ROOT volume.

**DEVELOPMENT=False (prod + staging with S3 creds):**
- Processing files live under `/tmp/scanning/{pk}/{reporter}/{vol}/{start}/` in whichever container produced them. MEDIA_ROOT is not used for processing files.
- Files are pushed to S3 (s3://bucket/processing/{pk}/...) at:
  - End of each pipeline (run_full_pipeline, run_reprocess, run_detect, run_generate_files, run_validate_with_bitonal).
  - Every time a user edit rewrites detections.json (single-file sync).
  - The original PDF is pushed synchronously by the upload view before the scan is queued.
- Files are deleted from /tmp/ only by the `cleanup_processing_tmp` management command, which the run_daemon scheduler invokes every `PROCESSING_TMP_CLEANUP_INTERVAL_SECONDS` (default 15 min). Directories idle for longer than `PROCESSING_TMP_TTL_HOURS` (default 24h) are removed.
- Files are re-downloaded from S3 to /tmp/ when:
  - A user opens the viewer (scan_process_view calls download_processing_files on every page load, size-idempotent).
  - A pipeline starts (_pull_processing_files_from_s3 runs at the top of every pipeline entry point, restoring /tmp/ if it was swept).
  - A lazy fallback in serve_opinionscan_pdf / serve_redacted_pdf fires if a request hits a missing file before the viewer's top-level hook could populate it.
- Approve (step 4) does a server-side s3.copy_object from `processing/{pk}/`... to `approved/{reporter}/{vol}/{start}/`. Nothing is re-uploaded from local disk, and source files stay in processing/.

Recovery: if an S3 upload silently failed (visible in Sentry), run `python manage.py reupload_scan_files <pk>` in the container that still has the local files in /tmp/ (within the TTL window).

scanning/migrations/0002_detection_issue_llmscan_pagedeletion_pageinsert_and_more.sql
grossir, line 1: do we really need the SQL file? This is used in Courtlistener to create new tables (which logical replication does not auto create)

Anyway, this file is out of sync. Some JSON fields are listed as TEXT fields, for example `margin_rects`

scanning/models.py
```diff
@@ -348,3 +757,171 @@ def __str__(self):
...
+        max_length=10,
+        choices=Severity.choices,
+        default=Severity.ERROR,
+    )
+    message = models.TextField()
+    metadata = models.TextField(
+        blank=True,
+        default="",
+        help_text="JSON for structured data (e.g. suppression info).",
+    )
```
grossir: if this is a JSON, should it be a JSON field?

grossir (commented):
Looks good. Just 2 comments @quevon24

scanning/migrations/0002_detection_issue_llmscan_pagedeletion_pageinsert_and_more.sql
quevon24: No, we don't need it, I'll remove the file

scanning/models.py
quevon24: yes, it should

grossir (approved):
LGTM!

PR #41, opened 2026-04-21
fix(storage): set default_acl to None

### LINKED ISSUE #40: collectstatic fails on deploy: AccessControlListNotSupported
The "Collect Static Assets" [step](https://github.com/freelawproject/scanning/actions/runs/24749052104/job/72408429564) of the deploy workflow is failing with

`botocore.exceptions.ClientError: An error occurred (AccessControlListNotSupported)` when calling the PutObject operation: The bucket does not allow ACLs. 

scanning/settings/third_party/aws.py sets AWS_DEFAULT_ACL = "private". django-storages forwards that value as an x-amz-acl header on every PutObject call. Our S3 buckets (com-freelawproject-scanning-storage, com-freelawproject-scanning-private-storage) were created after April 2023, when AWS changed the default Object Ownership for new buckets to Bucket owner enforced, which disables ACLs entirely. Any request carrying an ACL header — even "private" — is rejected with AccessControlListNotSupported.

Docs:

  - AWS — Controlling ownership of objects and disabling ACLs for your bucket: https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html — "Bucket owner enforced (default) – ACLs are disabled […] ACL-related requests fail with a 400 error with the error code AccessControlListNotSupported."                            
  - AWS News Blog — Heads-Up: Amazon S3 Security Changes Are Coming in April of 2023: https://aws.amazon.com/blogs/aws/heads-up-amazon-s3-security-changes-are-coming-in-april-of-2023/                                                                          
  - django-storages S3 backend docs (default_acl): https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html — "Default: None — the file will be private per Amazon's default."                                                                                                                                                       
                                                                                                                                                                             
For contrast, [courtlistener](https://github.com/freelawproject/courtlistener/blob/276d4ee3602debb9160b713c15a2bf870ad311c9/cl/lib/storage.py#L109) still uses default_acl = "private" on several storage classes; that works only because its buckets are legacy pre-2023 buckets with Object   Ownership still set to "Object writer" / "Bucket owner preferred", i.e. ACLs enabled.) 

So, the fix would be to set `AWS_DEFAULT_ACL = None`

I asked Claude how to confirm that objects will be still private:                                                                                                                                                                            

With Object Ownership set to Bucket owner enforced, ACLs no longer grant anything — access is controlled exclusively by bucket policy and IAM. Concretely:                 
   
  1. No ACL header is sent. django-storages omits x-amz-acl when default_acl is None, so PutObject is accepted without granting anyone extra permissions.                    
  2. Block Public Access remains on. New buckets have S3 Block Public Access enabled by default (same April 2023 change). To verify: S3 → bucket → Permissions → Block public
   access (bucket settings) — all four switches should be On.                                                                                                                
  3. Bucket policy grants nothing public. Confirm neither bucket has a Principal: "*" statement granting s3:GetObject without a scoped condition.                          
  4. Presigned URLs still work. We keep AWS_QUERYSTRING_AUTH = True, so URLs served by the app remain time-limited signed URLs — unchanged behavior.

grossir:
Solves #40

The scanning buckets do not support ACL. Privacy is controlled via bucket configuration. Setting ACL headers is causing errors. This PR removes the header from the Storage classes

We have confirmed the `PrivateS3Storage` remains private by bucket policy


quevon24 (approved):
All good, thanks for adding the docs in the issue explaining why it should have None

PR #44, opened 2026-04-23
fix(web): lazy-load blackletter to prevent scanning-web OOM at boot

### LINKED ISSUE #43: Lazy load blackletter imports in scanning.services to keep web pods under memory limit
After the blackletter integration, we are getting OOM errors for the web pod that [prevent](https://github.com/freelawproject/scanning/actions/runs/24754866957/job/72768012437) the deployment. 

We need to make heavy imports lazy so the web part never imports them

```
# failing pod
kubectl -n scanning describe pod scanning-web-6d6f5c9654-5z7z7


      Image ID:      docker.io/freelawproject/scanning@sha256:7ab54dc81f0966e8ed0ed893b29beb7bba6199bbaafc75c3b0bddcd6d251be0f                                                                                                           
      Args:                                                                                                                                                                  
        web-prod                                                                                                                                                             
      State:          Waiting                                                                                                                                                
        Reason:       CrashLoopBackOff                                                                                                                                       
      Last State:     Terminated                                                                                                                                             
        Reason:       OOMKilled                                                                                                                                              
        Exit Code:    137                                                                                                                                                    
        Started:      Thu, 23 Apr 2026 14:56:49 -0500                                                                                                                        
        Finished:     Thu, 23 Apr 2026 14:57:20 -0500                                                                                                                        
      Ready:          False                                                                                                                                                  
      Restart Count:  9                                                                                                                                                      
      Limits:                                                                                                                                                                
        memory:  500Mi                                                                                                                                                       
      Requests:                                                                                                                                                              
        memory:  500Mi                                                                                                                                                       

```

Imports happen here

```
  The heavy imports in the web path                                                                                                                                          
                                                                                                                                                                             
  All scanning-web workers (gunicorn ... --workers 4, one Python process each) walk this chain at boot:                                                                      
                  
  gunicorn worker                                                                                                                                                            
    └── scanning.asgi → django URLConf → scanning/urls.py
          ├── scanning/views_api.py     :11  import fitz              ← PyMuPDF                                                                                              
          └── scanning/views_process.py :8   import fitz                                                                                                                     
               :20  from scanning import services                                                                                                                            
                     │                                                                                                                                                       
                     ├── scanning/services.py :18  import fitz                                                                                                               
                     ├── scanning/services.py :19  from blackletter.analyze   → PIL                                                                                          
                     ├── scanning/services.py :22  from blackletter.api       → fitz                                                                                         
                     ├── scanning/services.py :31  from blackletter.margins   → fitz                                                                                         
                     ├── scanning/services.py :32  from blackletter.models    → (light)                                                                                      
                     ├── scanning/services.py :43  from blackletter.process   → fitz, ultralytics.YOLO                                                                       
                     ├── scanning/services.py :44  from blackletter.scanner   ★ cv2, numpy, PIL, ultralytics.YOLO, torch                                                     
                     └── scanning/services.py :45  from blackletter.validate  → .analyze, .process                                                                           

```


I asked Claude to estimate memory usage from the imports. Note that we have 4 workers, each one of them will make a copy of these

```
  Memory cost, module by module                                                                                                                                              
   
  Typical RSS delta after python -c "import X" on a Linux x86-64 CPython 3.12 process, CPU-only wheels from PyPI. These are measured orders of magnitude, not promises —     
  actual numbers vary by wheel variant (CPU-only vs. CUDA torch differs a lot) and glibc version:
                                                                                                                                                                             
  ┌─────────────────────────────────────┬─────────────────────────────────────────────────────────────────────┬─────────────────────────────────────────────────────────┐    
  │               Module                │                            What it loads                            │                        Added RSS                        │
  ├─────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤    
  │ Python + Django baseline            │ interpreter, Django, app code                                       │ ~70–100 MiB                                             │
  ├─────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ PIL / Pillow                        │ libjpeg, libpng, libtiff, openjpeg                                  │ ~10–15 MiB                                              │    
  ├─────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤    
  │ fitz (PyMuPDF)                      │ MuPDF, freetype                                                     │ ~25–40 MiB                                              │    
  ├─────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤    
  │ numpy                               │ OpenBLAS / MKL                                                      │ ~25–40 MiB                                              │
  ├─────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤    
  │ cv2 (opencv-python)                 │ ~50 MB of bundled .sos: imgproc, dnn, videoio, codecs               │ ~80–120 MiB                                             │
  ├─────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤    
  │ ultralytics import (no model        │ ultralytics itself is small; pulls pandas+matplotlib transitively   │ ~20–40 MiB                                              │
  │ instantiated)                       │ in some exporters                                                   │                                                         │    
  ├─────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ torch                               │ libtorch, ATen, MKL-DNN (~300 MB of shared libs mmapped)            │ ~250–400 MiB — CPU wheel ~220–280, CUDA wheel 350–450   │    
  │                                     │                                                                     │ even with no GPU                                        │
  ├─────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤    
  │ blackletter itself                  │ pure Python                                                         │ ~3–5 MiB                                                │
  └─────────────────────────────────────┴─────────────────────────────────────────────────────────────────────┴─────────────────────────────────────────────────────────┘    

  Summed for one worker after boot (CPU-only torch wheel, no YOLO weights loaded yet):                                                                                       
                  
  Django baseline              ~90 MiB                                                                                                                                       
  + PIL + fitz                 ~50 MiB
  + numpy                      ~30 MiB                                                                                                                                       
  + cv2                       ~100 MiB                                                                                                                                       
  + ultralytics               ~25 MiB
  + torch                     ~250 MiB                                                                                                                                       
  + blackletter python          ~5 MiB                                                                                                                                       
  ────────────────────────────────────
  ≈ 550 MiB per worker, idle, no weights loaded                                                                                                                              

```

grossir:
## Summary

- Move `from scanning import services` out of the top of `scanning/views_process.py` and into the one view that uses it (`recalculate`).
- Stops `scanning-web` gunicorn workers from transitively importing `blackletter.scanner` (and therefore `torch`, `cv2`, `ultralytics`, `numpy`) at boot.
- Unblocks rollouts that were failing with `OOMKilled` / `ProgressDeadlineExceeded` against the pod's 500Mi memory limit.

## Before vs. after (this branch)

| Metric | Before | After | Delta |
|---|---|---|---|
| Per-worker RSS after `import scanning.urls` | 427.4 MiB | 143.1 MiB | **−284 MiB (−66%)** |
| Heavy modules loaded by web path | torch, cv2, ultralytics, numpy, blackletter, fitz, PIL | fitz, numpy | — |
| Test suite (`manage.py test scanning.tests`) | 124 passed | 124 passed | no regression |

At `NUM_WORKERS=4` this brings pod-level RSS from ~1.7 GiB down to ~575 MiB. A follow-up bump of the k8s limit to ~768 MiB in the infra repo would give comfortable headroom for request spikes.

## Per-library attribution (isolated subprocess)

Numbers are from a fresh Python 3.13 process with CPU-only `torch==2.11.0+cpu`. "Added RSS" is the increase over the bare interpreter (14 MiB baseline).

| Import | Added RSS |
|---|---|
| django (module only, no `setup()`) | ~0 MiB |
| PIL (module only) | ~0 MiB |
| numpy | +16 MiB |
| fitz (PyMuPDF) | +34 MiB |
| cv2 | +42 MiB |
| torch | **+207 MiB** |
| ultralytics (transitively pulls torch + cv2 + numpy) | +237 MiB |
| blackletter (transitively pulls ultralytics + fitz) | +341 MiB |

## How these numbers were measured

1. Synced deps with `uv sync --all-extras` (produces the same CPU-only `torch==2.11.0+cpu` wheel used in the production image).
2. **Total web-path RSS:** a fresh Python subprocess calls `django.setup()`, imports `scanning.urls` (which pulls in every URLConf-level dependency a gunicorn worker would see), and reports `resource.getrusage(RUSAGE_SELF).ru_maxrss`. Ran the same script against `main` and against this branch.
3. **Per-library attribution:** for each library in the list, spawn a separate subprocess that performs only `import <lib>` and reports RSS. Isolation means the order of imports can't skew any single measurement.

## Test plan

- [x] `DEVELOPMENT=True DB_HOST=localhost DB_PORT=5434 DB_SSL_MODE=prefer python manage.py test scanning.tests -v 1` — 124 passed, 1 pre-existing skip.
- [ ] After merge, confirm the new ReplicaSet reaches Ready and the `Progressing=False/ProgressDeadlineExceeded` condition clears.
- [ ] `kubectl top pod -n scanning -l service=scanning-web` — expect ~400–600 MiB per pod, no OOMKills.
- [ ] Exercise the `recalculate` action on a scan in the process viewer to confirm the lazy import works end-to-end.

Refs #43

### DIFF
--- a/scanning/views_process.py
+++ b/scanning/views_process.py
@@ -17,7 +17,6 @@ from django.shortcuts import get_object_or_404, redirect, render
 from django.urls import reverse
 from django.views.decorators.http import require_POST
 
-from scanning import services
 from scanning.models import (
     CheckName,
     Detection,
@@ -538,6 +537,8 @@ def recalculate(request: HttpRequest, pk: int) -> HttpResponse:
     scan = get_object_or_404(Scan, pk=pk)
     if not scan.ocr_results:
         return redirect("scan_process", pk=pk)
+    from scanning import services
+
     services.recalculate_issues(scan)
     return redirect("scan_process", pk=pk)
 


quevon24 (commented):
Change looks good and the memory numbers speak for themselves. 

One thought: since the issue title targets `scanning.services` as the right fix location, would it be worth a follow-up to make the heavy imports lazy inside `services.py` itself (e.g. at the top of each function that needs them)? That way any future caller is automatically safe without needing to know about the eager-import side effect.

grossir:
@quevon24 let's get this one merged to see how the web container behaves. Then we can make the other changes which should be bigger

PR #45, opened 2026-04-24
feat: offload GPU steps (YOLO detection, OCR analysis) to RunPod Serverless

### LINKED ISSUE #42: Evaluate RunPod deployment strategy for the GPU-intensive scanning pipeline
## Context

The scanning pipeline has two GPU-heavy steps (`blackletter.detect` and `blackletter.analyze_pdf`) that we want to offload from the always-on CPU daemon onto rented GPU time. This issue compares the four realistic RunPod deployment strategies so we can pick one before committing more infrastructure work. Preliminary implementation targeting Serverless is already on the `feat/runpod-serverless` branch (scanning + `feat/ensure-weights` on blackletter).

### Option 1: RunPod Serverless

The daemon submits jobs to a Serverless endpoint. Workers spin up on demand, run one job, and die after an Idle Timeout window. The endpoint pulls a Docker image we push to a container registry. Input PDFs are handed to the worker as **presigned S3 GET URLs**, not raw AWS credentials.

**Advantages**
- Pay only when jobs are actually running. Zero cost while idle (assuming `Min Workers = 0`).
- No operational ownership: RunPod handles scheduling, capacity, retries, monitoring.
- Integration already implemented on the branch (`scanning/runpod_client.py`).
- Elastic: handles bursts by spawning concurrent workers.
- Single-toggle local fallback: daemon falls through to in-process blackletter when `RUNPOD_ENABLED=False`, so dev/CI are unaffected.
- **Tight security boundary on input access.** Presigned URLs are scoped to one specific S3 key, read-only, and expire after a configurable TTL. No AWS credentials are ever given to the worker. If a URL leaks (via logs, endpoint misconfiguration, registry compromise, etc.), the exposure is bounded to a single PDF for a limited time window, with no ability to list the bucket, write, or delete anything.

**Disadvantages**
- Queue-wait and cold-start times are unpredictable and can be multi-minute during busy hours.
- Each independent GPU call (detect, then analyze) risks its own cold start unless warm reuse kicks in.
- Presigned S3 URLs need a TTL long enough to survive queue + cold start (we are currently at 1 hour; max possible is 7 days).
- Pay-per-second rate on the 16 GB tier (A4000/A4500/RTX 2000 Ada/RTX 4000 Ada) is approximately `$0.00016/s ≈ $0.576/hr` of active worker time, which is more expensive per GPU-hour than renting a dedicated pod.

**Per-scan cost (measured on 1359-page volume, see Testing results below)**
- detect + analyze execution: ~$0.07
- Plus idle billing between/after calls (0 to $0.10): total ~$0.07 to ~$0.17 per scan.

**Requires**
- Docker image pushed to a registry (Docker Hub or GHCR).
- RunPod Serverless endpoint configured with GPU tier, Min/Max Workers, Idle Timeout.
- Env vars on the endpoint (`SENTRY_DSN_GPU`, `HANDLER_MAX_PAGES`, etc.).
- Daemon env vars (`RUNPOD_ENABLED`, `RUNPOD_ENDPOINT_ID`, `RUNPOD_API_KEY`, `RUNPOD_PRESIGNED_TTL`).
- S3 presigned GET URLs for the worker to read input PDFs.

---

### Option 2: Scheduled GPU pod (cron-based)

A dedicated GPU Pod is started on a fixed schedule (for example, 1 AM to 5 AM), processes whatever is in the queue, then stops. Controlled via `runpodctl` or the RunPod GraphQL API from a cron job.

**Advantages**
- GPU is guaranteed once the pod starts: no scheduler queue, no per-job cold start.
- Per-hour pricing is lower than Serverless: `$0.24 to $0.27/hr` for the 16 to 24 GB on-demand tier versus `$0.576/hr` equivalent on Serverless.
- Predictable: users know the processing window.
- Image and model weights persist on the pod's local disk between stops, so restarts are fast (seconds, not minutes).

**Disadvantages**
- Scans submitted outside the processing window sit in `QUEUED` until the next window. Unsuitable if users expect same-day feedback.
- We pay for the full window even if the queue is empty for part of it.
- Requires a different integration pattern on the daemon side: pods don't live behind `api.runpod.ai/v2/{endpoint}/run`, so `runpod_client.py` would need to talk directly to an HTTP server inside the pod (or we run the same `runpod.serverless.start` handler in the pod, see Option 4).
- A cron failure silently misses a processing window.

**Cost**
- Running 4 h/night on A4000 on-demand: `4 × $0.25 ≈ $1.00/day`.
- Running 8 h/night: ~$2.00/day.

**Requires**
- Cron job (on a scanning-owned server) or an equivalent scheduler.
- `runpodctl` or GraphQL API credentials for start/stop.
- Persistent pod definition.
- New integration in `runpod_client.py` to target a pod directly (or adoption of Option 4 to avoid this).

---

### Option 3: Daemon-triggered pod

The scanning daemon itself watches the `QUEUED` table; when non-empty, it calls the RunPod API to start a pod, dispatches scans to it, and stops the pod when the queue is empty.

**Advantages**
- Best cost profile of the pod options: pay only when there's actual work.
- GPU available on demand for each batch, with no queue-wait after startup.
- No reliance on a wall-clock schedule: responds to real load.
- Lower cost than always-on pods AND predictable availability.

**Disadvantages**
- The most code to own and debug. Needs state machine for pod lifecycle, health checks, startup timeouts, partial failures.
- First scan in a batch still pays the pod boot cost (2 to 5 min) before processing.
- Risk of pod-leak if the daemon crashes after starting the pod but before stopping it.
- Requires the same non-Serverless daemon integration as Option 2.

**Cost**
- Lower bound matches Serverless execution-only cost; upper bound is whatever the pod is active minus the per-hour-vs-per-second rate advantage. Break-even with Serverless at roughly 10 scans/day in a concentrated batch window.

**Requires**
- Pod lifecycle management code in the daemon.
- RunPod API credentials stored as a daemon env var.
- Failure-mode handling: pod failed to start, daemon crash, stuck pod, etc.
- New integration in `runpod_client.py` to target a pod directly (or Option 4).

---

### Option 4: GPU pod attached to the Serverless endpoint

A dedicated pod is attached to the Serverless endpoint as an "always-on worker." The daemon still calls `api.runpod.ai/v2/{endpoint}/run` exactly as in Option 1; RunPod routes work to the pod instead of a bursty Serverless pool. The pod can be scaled up/down on a schedule or kept 24/7.

**Advantages**
- Zero code change on the daemon side. Keeps everything we already built for Option 1.
- Zero cold-start ever: the pod is always warm.
- Same observability (RunPod dashboard, Sentry tagging, per-scan metrics).
- Can be combined with Option 2's cron scheduling to get Option 2 without the code rewrite.
- Same presigned-URL / no-AWS-creds security model as Option 1.

**Disadvantages**
- Most expensive if run 24/7: ~$6/day on A4000 on-demand, ~$180/month.
- Still subject to GPU scheduling at pod startup (but only on start, not per-job).
- Can't fully "turn off" without losing the pod-worker role unless we script the attach/detach. Scheduling helps here.

**Cost**
- 24/7: `$0.25/hr × 24 × 30 ≈ $180/month`.
- Scheduled 4 h/night: ~$30/month plus the existing Serverless bursty fallback outside that window.

**Requires**
- Everything Option 1 requires (image, endpoint, env vars, daemon wiring), already done.
- Additional pod attached to the endpoint via the RunPod console or API.
- Optional: cron job to start/stop the pod on a schedule.

---

## Comparison table

| Option | Upfront work | Ongoing cost (sporadic use) | Ongoing cost (40+ scans/day) | Cold start per job | Daemon code change |
|---|---|---|---|---|---|
| 1. Serverless | None (done) | ~$0.07 to $0.17/scan | ~$3 to $7/day | Likely | None |
| 2. Scheduled pod | Moderate (new integration) | ~$1 to $2/day fixed | ~$1 to $2/day fixed | Only on pod start | Yes |
| 3. Daemon-triggered pod | High (lifecycle code) | Low (pay only active) | Low | Only on first-of-batch | Yes |
| 4. Pod attached to endpoint | Low (zero code, just pod config) | $30/month (scheduled) to $180/month (24/7) | Flat | Never | None |

---

# Testing results from our Serverless implementation

All numbers below are from a single 1359-page case-reporter PDF processed against a Serverless endpoint with an RTX A4000/A4500 tier (16 GB VRAM) on RunPod, image pushed to a personal Docker Hub repo.

## Detect

| Field | Value |
|---|---|
| `status` | `COMPLETED` |
| `delayTime` (queue + worker boot) | 1,325,960 ms (~22 min) |
| `executionTime` (RunPod's view) | 317,131 ms (~5.3 min) |
| `duration_ms` (handler's view) | 297,808 ms (~5.0 min) |
| `worker_boot_ms` (cold-start cost) | 3,313 ms (~3.3 s) |
| `worker_uptime_ms` at job start | 318,311 ms |
| `gpu_available` | `true` |
| `page_count` | 1,359 |
| Response payload | 2.6 MB |

**Per-YOLO-model durations:**

| Model | Duration | Per-page |
|---|---|---|
| `small` (21 MB weights) | 84,557 ms (~84.6 s) | 62 ms/page |
| `medium` (49 MB weights) | 103,082 ms (~103.1 s) | 76 ms/page |
| `large` (148 MB weights) | 109,465 ms (~109.5 s) | 81 ms/page |

**Detection quality**
- Total detections: 10,057 across 1359 pages (~7.4/page).
- Every page has at least one detection.
- Multi-model agreement: 57% of detections found by all 3 models, 27% by 2, 9% by 1 only.

## Analyze (PaddleOCR + YOLO large)

| Field | Value |
|---|---|
| `status` | `COMPLETED` |
| `delayTime` | 567,055 ms (~9.5 min) |
| `executionTime` | 150,276 ms (~2.5 min) |
| `duration_ms` | 137,953 ms (~2.3 min) |
| `worker_boot_ms` | 3,258 ms (~3.3 s) |
| `gpu_available` | `true` |
| Response payload | 873 KB |

**Page-number detection quality**
- 1359/1359 pages (100%) received a detected page number.
- 1356 of those located via YOLO's page-number box (direct read).
- 3 via header-side fallback when YOLO missed.
- Zero anomalous jumps in the numeric sequence.

## Warm reuse

Both jobs in the same test run were dispatched to the **same worker** (`sywqhmmostejwk`). Analyze finished, the worker stayed `Idle` for the 5-minute Idle Timeout window, then the detect submission picked it up without a fresh cold start. This is the expected behavior for a single scan going through `run_full_pipeline` and confirms the client + endpoint wiring.

## Cost per 1359-page scan (at \$0.00016/s)

- Detect execution: 298 s × \$0.00016 = **\$0.048**
- Analyze execution: 138 s × \$0.00016 = **\$0.022**
- Cold boot per GPU call: ~3 s, ~\$0.001
- Idle between detect and analyze (up to 300 s): **\$0 to \$0.048**
- Idle after last call (up to 300 s until Idle Timeout): **\$0 to \$0.048**

**Total: ~\$0.07 to ~\$0.17 per full-pipeline scan**, depending on how long warm-idle billing runs.

## Issues encountered during Serverless implementation

All resolved on the branch, documented here for completeness.

1. **paddlepaddle-gpu vs torch CUDA conflict.** On long YOLO runs (1000+ pages), Paddle's idle CUDA allocator fired a cleanup pass while torch was driving the GPU, producing `cudaErrorInitializationError` and SIGABRT. Fix: removed PaddleOCR preload from `_preload()`. PaddleOCR now loads lazily on first analyze call (blackletter already does this). Detect-only workers never load Paddle at all.
2. **Worker being killed between jobs.** Initially thought to be a RunPod SDK bug; actually a misconfigured `Idle Timeout` of **5 seconds** instead of 300 s. Fix: set to 300 s in the endpoint config. Warm reuse now works.
3. **`Failed to get job. | missing field(s): id or input`** on `/runsync` submissions. Workers boot successfully, SDK then crashes on job dispatch. Fix: use `/run` (async + poll) from the daemon. `runpod_client.py` uses `/run` exclusively.
4. **Docker image originally 37.6 GB** due to uv's wheel cache being baked into the layer. Fix: `--mount=type=cache,target=/root/.cache/uv` + `uv sync --no-cache` in the Dockerfile. Now 22 GB uncompressed / 7.09 GB compressed (what Docker Hub stores and RunPod pulls).
5. **`libcuda.so.1` missing at build time**, breaks import of `paddlepaddle-gpu`. Fix: multi-stage `COPY --from=nvidia/cuda:*-devel-*` of the stub library. At runtime, RunPod's nvidia-container-toolkit bind-mounts the real `libcuda.so.1` over the stub.

## Open question: presigned URL TTL vs queue wait

Our current implementation generates presigned GET URLs for the worker to read input PDFs (no AWS credentials are given to the worker). Presigned URLs have a TTL; if RunPod's queue + cold-start exceeds that TTL, the worker gets a 403 when it finally tries to fetch the PDF.

Observed queue + boot times in our testing have reached 22 minutes. The current TTL is 1 hour, which is fine so far, but busy-hour scheduling could push it higher.

Mitigation options (to decide alongside the deployment option):

- **Bump TTL to the AWS SigV4 maximum of 7 days.** Zero code change, very low security impact (URL is still read-only, scoped to one specific S3 key, unguessable signature).
- **Give the worker scoped read-only IAM credentials instead of presigned URLs.** More code, slightly tighter security (revokable, auditable via CloudTrail).
- **Keep 1-hour TTL and add 403-retry logic on the daemon side.** Most code, only defends against URL expiry specifically.

quevon24:
## What this does

Introduces a RunPod Serverless integration that offloads the two GPU-intensive pipeline steps — YOLO detection and PaddleOCR analysis — to on-demand GPU workers, while keeping the CPU-bound steps in the daemon where they belong.

### What runs on GPU (RunPod)

- **YOLO detection** (`detect` action): runs all three models (small, medium, large) on the bitonal PDF and returns the merged detection list
- **PaddleOCR analysis** (`analyze` action): runs page-number extraction and validation on the original PDF

Both steps previously ran on the daemon's CPU, which made them significantly slower and blocked the daemon for the duration.

### What stays on CPU (daemon)

- **Bitonal conversion**: fast, purely CPU-bound image processing — no GPU benefit
- **Tesseract OCR**: `ocrmypdf` parallelises across all available cores and produces good results on CPU; adding a GPU worker for this step would add latency without meaningful throughput gain
- **Opinion pairing, redaction rects, file generation**: lightweight computation that finishes in seconds and has no GPU-accelerated equivalent

### Other changes

- **Transient retry cap**: RunPod failures (e.g. worker scheduled without a GPU) are automatically retried up to `RUNPOD_MAX_TRANSIENT_RETRIES` times (default 5) using an atomic `CASE/WHEN` update. Scans that exhaust retries land in a new `error_max_retries` status.
- **Admin tooling**: staff can filter and re-queue retry-capped scans from the Django admin. A banner on the scan list alerts all users when capped scans exist.
- **Observability**: RunPod job logs include pod execution time, action duration, and per-model YOLO timings. Sentry tags carry the scan PK and RunPod job ID for every captured exception.
- **Local fallback**: `RUNPOD_ENABLED=False` (the default in dev) runs all steps in-process, so local development and tests require no RunPod credentials.

### Daemon environment variables

| Variable | Required | Default | Description |
|---|---|---|---|
| `RUNPOD_ENABLED` | No | `False` | Master switch; set to `True` in production |
| `RUNPOD_ENDPOINT_ID` | Yes (prod) | `""` | Serverless endpoint ID from the RunPod console |
| `RUNPOD_API_KEY` | Yes (prod) | `""` | RunPod API key (treat as secret, never log) |
| `RUNPOD_REQUEST_TIMEOUT` | No | `1800` | Max seconds to wait for a job (submit + queue + execution) |
| `RUNPOD_MAX_RETRIES` | No | `2` | Retries on transport errors when submitting |
| `RUNPOD_MAX_TRANSIENT_RETRIES` | No | `5` | Retry cap before a scan is escalated to `error_max_retries` |
| `RUNPOD_PRESIGNED_TTL` | No | `86400` | Lifetime in seconds of the presigned S3 URL handed to the worker |

### Worker image

The GPU worker is a separate Docker image defined in `scanning/runpod/`. It contains only inference code and baked-in model weights (YOLO small/medium/large + PaddleOCR PP-OCRv5) — no Django, no database client, no AWS credentials. Full setup and troubleshooting details are in `scanning/runpod/README.md`.

Once this PR is merged, a follow-up GitHub Action should be added to automatically build and push the worker image to the container registry whenever `scanning/runpod/` changes.

The RunPod endpoint should be configured with the following env vars (set on the endpoint, not on the daemon):

| Variable | Description |
|---|---|
| `SENTRY_DSN_GPU` | Separate Sentry project for the GPU worker |
| `SENTRY_ENV` | `prod`, `staging`, etc. |
| `HANDLER_MAX_PAGES` | Reject PDFs above this page count (default 5000) |
| `HANDLER_DOWNLOAD_TIMEOUT` | Seconds to wait for the presigned URL download (default 300) |

Paired with @quevon24 and [Claude Code](https://claude.com/claude-code)


grossir:
Related to the deployment of this PR, once it's merged

@quevon24 can you create an issue on the infra repo about what env variables we will need to set?

Also, what needs to be configured on the runpod endpoint config itself. From the README. I guess you will have to create the Sentry project, etc

```
7. **Env vars** (set in the endpoint config, NOT in the image):
   - `SENTRY_DSN_GPU` — separate Sentry project from Django.
   - `SENTRY_ENV` — `prod` / `staging` / etc.
   - `GIT_SHA` — optional, used as the Sentry release tag.
   - `HANDLER_MAX_PAGES` — reject PDFs above this many pages
     (default 5000).
   - `HANDLER_DOWNLOAD_TIMEOUT` — seconds to wait for the presigned
     GET URL download (default 300).
```

scanning/services.py
```diff
@@ -1349,6 +1478,7 @@ def run_full_pipeline(scan_pk: int) -> None:
     :param scan_pk: Primary key of the scan to process.
     """
     django.db.connections.close_all()
+    Scan.objects.filter(pk=scan_pk).update(retry_count=0)
```
grossir: each one of the `run_*` functions sets `Scan.objects.filter(pk=scan_pk).update(retry_count=0)` . Is this desired?

- run_full_pipeline
- run_validate_with_bitonal 
- run_reprocess
- run_generate_files
- run_detect

If each step is independent, as they seem to be, maybe it is desired

grossir: I think this is a bug. If processing fails, `handle_pipeline_exception` is called, which updates the retry_count and sets the status back to QUEUED. But, any daemon entry point over queued objects will set retry_count back to 0

grossir (commented):
I think there is a bug with retry_count logic @quevon24

scanning/services.py
quevon24: Good catch! My reasoning was that `retry_count` should represent failures for the current pipeline invocation, so resetting it at the entry point would give each intentional run a clean slate. What I missed is that automatic retries after a transient failure also go through the same QUEUED -> PROCESSING path and reach the same run_* entry points, so there's no way to distinguish a fresh run from a retry at that level. So the reset was triggered on every attempt, making the limit unreachable

scanning/runpod_client.py
```diff
@@ -0,0 +1,594 @@
...
+    base_url = f"https://api.runpod.ai/v2/{endpoint}"
+    headers = {"Authorization": f"Bearer {api_key}"}
+    body = {
+        "input": {"action": action, "scan_pk": scan.pk, **payload},
+    }
+
+    deadline = time.monotonic() + int(
+        getattr(settings, "RUNPOD_REQUEST_TIMEOUT", 900)
+    )
+    max_retries = int(getattr(settings, "RUNPOD_MAX_RETRIES", 2))
```
grossir: these are basically overriding the defaults on settings.rundpod

grossir (commented):
@quevon24 this looks good. Only a small comment on settings override

grossir (approved):
LGTM

PR #48, opened 2026-04-29
fix(entrypoint): Disables Gunicorn control socket to prevent permission errors

### LINKED ISSUE #47: gunicorn control-server errors
As part of ongoing efforts to improve our deployment settings, specifically to keep pods running longer and cut down on updown alerts, we’ve been monitoring pods more closely. During this, we noticed the following error:

```
Control server error: [Errno 13] [Errno 13] Permission denied: '/var/www'
```

This error looks related to a change introduced in version 25, which enables the Gunicorn [control socket](https://gunicorn.org/reference/settings/#control_socket) by default. By default, the socket is created at `$XDG_RUNTIME_DIR/gunicorn.ctl` or `$HOME/.gunicorn/gunicorn.ctl`. In our pods, $HOME resolves to /var/www.

Since our deployment runs as an unprivileged user, it doesn’t have permission to write to `/var`, which leads to the error.

The control socket is used for runtime management of Gunicorn (e.g., inspecting workers, adjusting worker count, graceful reloads) via the gunicornc CLI. However, we’re not currently using this feature.

Since we’re not using this feature, we can probably just disable the control socket to avoid these errors in production.

ERosendo:
## Summary

Adds `--no-control-socket` to the Gunicorn production command to disable the control socket feature introduced in Gunicorn 25

## Background

Gunicorn 25 enables a control socket by default, writing to `$XDG_RUNTIME_DIR/gunicorn.ctl` or `$HOME/.gunicorn/gunicorn.ctl`. In our pods, `$HOME` resolves to `/var/www`, which the `www-data` user cannot write to, producing repeated arbiter errors:

```
Control server error: [Errno 13] Permission denied: '/var/www'
```

Since we don't use `gunicornc` for runtime worker management, disabling the socket is the right fix.

Closes #47

### DIFF
--- a/docker/django/docker-entrypoint.sh
+++ b/docker/django/docker-entrypoint.sh
@@ -32,6 +32,7 @@ case "$1" in
         --group www-data \
         --workers ${NUM_WORKERS:-4} \
         --worker-class scanning.workers.UvicornWorker \
+        --no-control-socket \
         --timeout 180 \
         --max-requests ${MAX_REQUESTS:-2500} \
         --max-requests-jitter 100 \


mlissner:
Went ahead and did this/deployed it for the wiki.

quevon24 (approved):
Approved! Looks good. Thanks for the explanation.

PR #50, opened 2026-04-29
feat(scanning): detection box resize, move, and page-container width fix

quevon24:
## Summary

- Add resize handles (8-point) to existing detection bounding boxes in the process viewer step 2. Double-click a box to select it, then drag a handle to resize.
- Add drag-to-move on selected detection boxes. After double-clicking to select, drag the box body to reposition it.
- Add drag-to-move on selected redaction overlay boxes. After double-clicking to select, drag the box body to reposition it.
- New `update-detection` API endpoint persists detection bbox changes to the DB and `detections.json`.
- Fix blank space appearing next to whiteout blocks when the browser inspector is open: constrain `page-container` width to match the rendered canvas width so the page-label buttons cannot make the container wider than the canvas.
- Add tests for the `update_detection` endpoint covering the happy path, matching by `label_id`, and missing `detections.json`.

### DIFF
--- a/scanning/static/scanning/viewer_step2.js
+++ b/scanning/static/scanning/viewer_step2.js
@@ -749,6 +749,10 @@ document.addEventListener('DOMContentLoaded', function () {
             box.style.top = (d.bbox[1] * sy) + 'px';
             box.style.width = ((d.bbox[2] - d.bbox[0]) * sx) + 'px';
             box.style.height = ((d.bbox[3] - d.bbox[1]) * sy) + 'px';
+            box.dataset.sx = sx;
+            box.dataset.sy = sy;
+            box.dataset.imgWidth = imgW;
+            box.dataset.imgHeight = imgH;
             box.style.borderColor = color;
             box.title = d.label + ' (' + d.confidence + ')';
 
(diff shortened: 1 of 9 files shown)


scanning/views_api.py
```diff
@@ -732,6 +732,73 @@ def delete_detection(request: HttpRequest, pk: int) -> JsonResponse:
...
+    old_bbox = data["old_bbox"]
+    new_bbox = data["new_bbox"]
+    db_filter = dict(
+        scan=scan,
+        page_index=page_index,
+        x0__gte=old_bbox[0] - 15,
+        x0__lte=old_bbox[0] + 15,
+        y0__gte=old_bbox[1] - 15,
+        y0__lte=old_bbox[1] + 15,
+    )
```
grossir: instead of doing this, which may be a bit error prone, and probably slower

Why not send the Detection.id in serve_detections, and then use the returned id for editions?

https://github.com/freelawproject/scanning/blob/814416a12c05caaa84adecbc07b996f4329fc217/scanning/views_api.py#L74-L98

scanning/views_api.py
```diff
@@ -732,6 +732,73 @@ def delete_detection(request: HttpRequest, pk: int) -> JsonResponse:
...
+        db_filter["label"] = label
+    elif label_id is not None:
+        db_filter["label_id"] = label_id
+    count = Detection.objects.filter(**db_filter).update(
+        x0=new_bbox[0], y0=new_bbox[1], x1=new_bbox[2], y1=new_bbox[3]
+    )
+    output_base = Path(scan.output_dir)
+    det_path = find_json_file(output_base, "detections.json")
+    if not det_path:
+        return JsonResponse({"error": "No detections.json"}, status=404)
```
grossir: the DB update always succeeds, right? If that happens, we should ensure the S3 files and the local files are updated too, right?

So, if the JSON file is not currently on disk, it should be downloaded to S3? And then  `_sync_detections_to_disk` called?

scanning/static/scanning/viewer_step2.js
```diff
@@ -1712,13 +1767,143 @@ document.addEventListener('DOMContentLoaded', function () {
...
+                        method: 'POST',
+                        headers: {'Content-Type': 'application/json', 'X-CSRFToken': csrfToken},
+                        body: JSON.stringify({
+                            page_index: det.page_index,
+                            label:      det.label,
+                            label_id:   det.label_id,
+                            old_bbox:   oldBbox,
+                            new_bbox:   newBbox,
+                        }),
+                    }).then(function(r) { return r.json(); }).then(function(data) {
```
grossir: this may actually fail with 404, but there is no error catching? So the user wouldn't know if it failed

grossir (commented):
Some comments to be addressed

quevon24:
Addressed all three points:
- serve_detections now includes the DB id in each detection, and the detection_id approach was also applied to delete_detection and approve_detection, all three now use exact pk lookup instead of fuzzy bbox matching
- output_dir is created before calling _sync_detections_to_disk, which also handles the S3 upload
- Added .catch error handling to all detection fetch calls in the JS, errors will be logged to the console

quevon24:
I found a bug while testing bbox editing: clicking or double-clicking a detection, redaction, or margin box was triggering save endpoints (update-detection/, save-redaction-rect/, save-margin-rect/) even without any changes. 

Also added a stacking dismissable toast to display errors when detection actions (delete, approve, move, resize) fail on the server, while keeping the existing console.error logging.

scanning/tests/test_views.py
```diff
@@ -1033,4 +1033,254 @@ def test_invokes_pull_helper(self):
 
             run_full_pipeline(scan.pk)
 
-        mock_pull.assert_called_once_with(scan.pk)
```
grossir, line 1036: why delete this assertion? seems unrelated to the current changes

scanning/views_api.py
```diff
@@ -683,55 +684,63 @@ def remove_flag(request: HttpRequest, pk: int, flag_id: int) -> JsonResponse:
...
+        return data
+    detection_id = data["detection_id"]
+    new_bbox = data["new_bbox"]
+    count = Detection.objects.filter(pk=detection_id, scan=scan).update(
+        x0=new_bbox[0],
+        y0=new_bbox[1],
+        x1=new_bbox[2],
+        y1=new_bbox[3],
+    )
+    output_dir = Path(scan.output_dir)
```
grossir, line 746: there should be a check around here to see if count == 0. 
If that happens, the response should not be a 200 OK. 
It may happen if something went wrong somewhere (maybe 2 concurrent editions of a scan?)

grossir (commented):
Looking good. Some smaller comments

grossir (approved):
LGTM

PR #51, opened 2026-04-30
fix(runpod): pin blackletter>=0.0.9 and fix PADDLE_PDX_CACHE_HOME

quevon24:
## Summary

- Pins `blackletter>=0.0.9` in both `pyproject.toml` (main project) and `scanning/runpod/pyproject.toml`, replacing the git branch pin
- Fixes the PaddleX model cache directory: `PADDLEX_HOME` is not a recognized env var in paddlex 3.x — the correct one is [`PADDLE_PDX_CACHE_HOME`](https://github.com/PaddlePaddle/PaddleX/blob/release/3.5/paddlex/utils/cache.py#L29C53-L29C70). With the wrong var set, models were silently falling back to `/root/.paddlex` instead of `/opt/paddlex`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,9 +29,7 @@ dependencies = [
   "gunicorn",
   "pillow",
   "psycopg[binary,pool]",
-  # TODO: flip back to a PyPI pin once blackletter releases a version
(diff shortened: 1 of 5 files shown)


grossir (approved):
LGTM

PR #53, opened 2026-05-01
fix(runpod): exit CPU-only workers before accepting jobs

quevon24:
## Problem

When RunPod's scheduler assigned a serverless worker to a CPU-only host, `_preload()` detected no GPU and returned early, but the worker process kept running. It would accept jobs, immediately return `error_code=NO_GPU`, and wait for the next one. Since RunPod bills from the moment a worker boots until it is terminated, this wasted money on workers that could never do useful work.

## Fix

**Fitness check in `handler.py`**

A `@runpod.serverless.register_fitness_check` function now checks `_CUDA_AVAILABLE` at worker startup. If no GPU is present, the worker exits with `sys.exit(1)` before RunPod's heartbeat starts, meaning it is never added to the available pool and no jobs are ever routed to it. Queued jobs are picked up automatically by healthy GPU workers with no daemon changes needed.

The fitness check runs before the heartbeat and job loop (see [Worker Fitness Checks](https://github.com/runpod/runpod-python/blob/main/docs/serverless/worker_fitness_checks.md)), so the worker never bills for job time, only the cold-start cost up to the check.

The handler still returns `error_code=NO_GPU` as a belt-and-suspenders fallback for the unlikely case where CUDA disappears after startup. `NO_GPU` remains in `_TRANSIENT_ERROR_CODES` so the daemon re-queues the scan normally.

Also included: a cosmetic fix to the terminal-status error message (previously produced `"RunPod job X CANCELLED: CANCELLED"`) and README instructions for deploying a new image version to an existing endpoint.

### DIFF
--- a/scanning/runpod/README.md
+++ b/scanning/runpod/README.md
@@ -104,6 +104,39 @@ docker tag blackletter-gpu-worker:local \
 docker push ghcr.io/freelawproject/blackletter-gpu-worker:<tag>
 ```
 
+## Releasing a new image version
+
+Pushing a new image tag to the registry does **not** automatically update
+the running endpoint. You must create a new release in the RunPod console
+so workers boot from the new image:
+
+1. Go to the **Serverless** page and open your endpoint.
+2. Click **Manage**.
+3. Select **New Release**.
+4. In the **Container Image** field, paste the full new tag, e.g.
+   `freelawproject/blackletter-gpu-worker:ff73140`.
+5. Save. Existing warm workers drain their current jobs; new workers
+   start from the updated image.
+
+### Why we pin a commit SHA instead of using `:latest`
+
+Using `:latest` is unreliable for serverless deployments for three reasons:
+
+- **Caching conflicts.** RunPod caches images on each node for faster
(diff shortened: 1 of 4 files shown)


scanning/runpod/README.md
```diff
@@ -104,6 +104,39 @@ docker tag blackletter-gpu-worker:local \
...
+the running endpoint. You must create a new release in the RunPod console
+so workers boot from the new image:
+
+1. Go to the **Serverless** page and open your endpoint.
+2. Click **Manage**.
+3. Select **New Release**.
+4. In the **Container Image** field, paste the full new tag, e.g.
+   `freelawproject/blackletter-gpu-worker:ff73140`.
+5. Save. Existing warm workers drain their current jobs; new workers
+   start from the updated image.
```
grossir, line 119: it seems this could be done programtically? 

It seems we could use the API to create a template ponting to the latest commit, then PATCH the endpoint with the new template

https://lucaschmid.net/blog/runpod-deploy/

grossir (approved):
Looks good, but I think we could integrate the runpod image update to our CI

PR #54, opened 2026-05-02
fix(runpod): handle FAILED job responses by error source

quevon24:
The RunPod SDK pops `error` from the handler's return dict and places it at the top level of the result payload. This means handler errors (bad input, no GPU, unknown action) arrive as `FAILED` jobs with `error_code` inside `output`, not as `COMPLETED` jobs with `error` in `output`.

Previously the client had a dead code path checking `"error" in output` on COMPLETED responses (never triggered), and treated all FAILED/TIMED_OUT/CANCELLED as terminal errors.

**New behavior:**
- `FAILED` + `output.error_code` present: transient if `NO_GPU`, terminal otherwise
- `FAILED` + no `output`: transient (RunPod platform failure, e.g. "job timed out after 1 retries", worker crash)
- `TIMED_OUT` / `CANCELLED`: terminal

### DIFF
--- a/scanning/runpod/handler.py
+++ b/scanning/runpod/handler.py
@@ -461,10 +461,11 @@ def handler(job: dict) -> dict:
         (cold-start cost of this worker process, constant per
         worker), ``worker_uptime_ms`` (ms since preload finished, at
         job start), and ``gpu_available`` (whether torch.cuda saw a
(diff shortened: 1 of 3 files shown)


grossir (approved):
Looks good

Issue #59, opened 2026-05-04
Document difference between `Scan.page_count` and `Scan.number_of_pages`


grossir:
We have 2 fields on the Scan model tracking a single characteristic 

`number_of_pages` is never populated? It is used for validation, though; 

https://github.com/freelawproject/scanning/blob/92fd24218bca567c3a3787ebc6cfa382a330b8d7/scanning/models.py#L538-L541
https://github.com/freelawproject/scanning/blob/92fd24218bca567c3a3787ebc6cfa382a330b8d7/scanning/models.py#L424-L428

Error apparent here where all "Pages" values are None. On the detail view, each scanned volume has an actual page count, from `Scan.page_count`

<img width="1326" height="427" alt="Image" src="https://github.com/user-attachments/assets/693854c3-d553-4937-befb-e69304698689" />

grossir:
After talking with Bill about this issue, the field is not actually duplicated and should not be consolidates. This is a documentation issue. The Scan.number_of_pages field will be populated by the scanner when filling the upload form. Since we uploaded these initial volumes without following the procedure, we get the None values on the table. May be useful to use Scan.page_count as a default for that case

- page_count: the actual scanned PDF file page count
- number_of_pages: should track the "canonical" number of pages, that doesn't account for withdrawn opinions or special cases

From Bill
> Sometimes the number of pages differs because … i saw once 1 thru 1390 and 1390A and 1390B
> others may have page ranges … so it may be 1- 1290 but pages 700 - 777 wont exist
> so the number of pages is what 1283.  but pages 1 thru 1290

grossir:
This actually brings a modelling question. Seems like `number_of_pages` should be a property of Volume. Taking a note.. let's not overcomplicate the solution to this

PR #62, opened 2026-05-05
fix(Scan.number_of_pages): consolidate field into Scan.page_count

grossir:
Solves #59

Replaces all number_of_pages by page_count on the Scan model

No need for a data migration since all scans we have have a page_count value; and no number_of_pages values


grossir:
Closing because the fields are not actually duplicated. See related issue for more detail #59

PR #63, opened 2026-05-08
feat(viewer): add zoom controls to processing pdf viewer

### LINKED ISSUE #57: Add zoom controls to the pdf viewer
This should be a small change. On my viewport it is hard to read the opinion itself. It may be because I have a big monitor, but I imagine other viewports may run into this problem. Won't post a screenshot but you can check on the scanning portal itself

quevon24:
## Summary
Closes #57.

Adds a zoom toolbar (−, %, +, Reset) to the processing PDF viewer (`scan_process.html`, used in steps 1, 2, and 3). Range 0.5x – 3x, step 0.1. Step 1 persists its zoom level independently from steps 2/3, since reviewing page numbers and reviewing opinions benefit from different zooms.

## How it works
The toolbar lives at the top-left of the viewer panel and stays visible when you scroll, even horizontally at high zoom. Each PDF page is scaled visually as a whole, so detection boxes, redactions, and image overlays all stay aligned with the page automatically. Drawing a new redaction or detection, and dragging or resizing existing ones, works the same at any zoom level.

pdf.js rasterizes each page once into a canvas at a fixed resolution, so simply CSS-scaling that bitmap to a higher zoom looks really blurry above the original render scale (the same effect as zooming into a JPEG). To avoid this, when the zoom level changes the viewer re-rasterizes visible pages at the new resolution so text stays sharp. Pages outside the viewport are dropped and re-rendered fresh when scrolled back into view, which keeps memory bounded at high zoom on long documents. Rapid zoom clicks are debounced and any in-flight render is cancelled when zoom changes again, the same pattern pdf.js's own demo viewer uses.

The page header (PDF p.n + action buttons) also got a small layout tweak: at narrow widths (low zoom or small monitors) the buttons wrap below the page identifier instead of overflowing the page width.

### DIFF
--- a/scanning/static/scanning/checker.css
+++ b/scanning/static/scanning/checker.css
@@ -376,6 +376,8 @@ a.doc-name:hover { text-decoration: underline; }
     display: flex;
     align-items: center;
     justify-content: space-between;
+    flex-wrap: wrap;
+    row-gap: 4px;
     border-bottom: 1px solid #eee;
 }
 
--- a/scanning/static/scanning/checker.css
+++ b/scanning/static/scanning/checker.css
@@ -282,9 +282,52 @@ a.doc-name:hover { text-decoration: underline; }
 /* PDF Viewer */
 .viewer-panel {
     flex: 1;
-    overflow-y: auto;
+    overflow: auto;
     background: #525659;
     padding: 20px;
+    position: relative;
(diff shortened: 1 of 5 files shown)


grossir (approved):
LGTM!

PR #64, opened 2026-05-08
fix(scan_list): default to page_count when number_of_pages is null

### LINKED ISSUE #59: Document difference between `Scan.page_count` and `Scan.number_of_pages`

grossir:
Solves #59

Add help_text to Scan.number_of_pages and Scan.page_count to make their difference understandable

Makes the scan_list "Pages" default to page_count when number_of_pages is null


quevon24 (approved):
LGTM. The help text additions make the distinction between number_of_pages and page_count much clearer.

PR #66, opened 2026-05-12
feat(scanning): keep Volume.queue_status in sync with scan lifecycle

### LINKED ISSUE #58: Queue status is not updated when the scan process is completed
1. The approved files should show `queue_status = completed` once the reporter has been approved
2. We should improve the lifecycle of the queue status field
3. we should differentiate the badges in this status column somehow. We have Scan.queue_status and Scan.status and Scan.s3_uploaded badge. At least we should include a tooltip over the badges that explains them a bit further

https://github.com/freelawproject/scanning/blob/92fd24218bca567c3a3787ebc6cfa382a330b8d7/scanning/models.py#L106-L112

Screenshot of current state:
Image

quevon24:
## Summary
Addresses issue #58:

- **Approval → `complete`**: `Volume.queue_status` now advances automatically when its scans are approved, instead of staying at `scanning` / `scanned` indefinitely.
- **Lifecycle**: A single helper in `services.py` derives `queue_status` from observable scan state and is called from every touchpoint (upload, manual review, auto-approval, admin deletion). Manual `unavailable` is preserved as a curator override.
- **Badges**: Hover tooltips on the three Status-column badges on `/queue/` (`Volume.queue_status`, `Scan.status`, `S3`) so it's clear what each one represents.

14 new tests; full suite passing.

Closes #58

### DIFF
--- a/scanning/admin.py
+++ b/scanning/admin.py
@@ -11,6 +11,7 @@ from scanning.models import (
     Status,
     Volume,
 )
+from scanning.services import refresh_volume_queue_status
 
 
 class RetryCapFilter(admin.SimpleListFilter):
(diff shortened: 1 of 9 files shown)


grossir (commented):
Looks good! I think it's missing a call to `refresh_volume_queue_status` here

https://github.com/freelawproject/scanning/blob/994ec90195681d5dadec8591311c112e0edf9250/scanning/views_api.py#L410-L413

grossir (approved):
Looks good!

PR #69, opened 2026-05-13
perf(views): revalidate opinion pdf responses with etag

### LINKED ISSUE #56: Cache viewer file requests in the frontend
### Current behavior

When alternating Redacted / Original files, every click will trigger a request, even if we have already requested the file from the server. See the original  and redacted requests 

Image

### Expected behavior

Case 1: viewing the files

If we are only "viewing" the files (not reviewing / editing the redactions and other stuff) then the request should happen a single time, and be cached in the frontend

Case 2: editing the file

If an edition happens, the file should load each time changes are saved

quevon24:
## Summary

- Closes #56.
- Adds `ETag` (file mtime + size), `Last-Modified`, and `Cache-Control: private, no-cache` to `serve_opinionscan_pdf`. Toggling between Redacted / Original / Masked variants in the step 3 viewer revalidates against the server and returns 304 when the file is unchanged, so the multi-MB PDF body is no longer re-downloaded on every click.
- Edits invalidate the cache automatically: in-place rewrites change the file's mtime/size (so the ETag changes), and Regenerate Files creates new `OpinionScan` rows with new PKs (so the URL changes). No `max-age` window, no implicit "all edit paths must bust the URL" contract.

### DIFF
--- a/scanning/tests/test_views.py
+++ b/scanning/tests/test_views.py
@@ -1087,6 +1087,44 @@ class TestServeOpinionPdfLazyPull(ScanningTestCase):
         mock_pull.assert_called_once()
 
 
+class TestServeOpinionPdfCaching(ScanningTestCase):
+    """serve_opinionscan_pdf sets cache headers and honors revalidation."""
+
+    def test_response_carries_cache_headers(self):
+        user = self.make_user()
+        self.client.force_login(user)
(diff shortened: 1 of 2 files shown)


grossir (approved):
Looks good @quevon24 .

PR #70, opened 2026-05-13
feat(runpod): allow local end-to-end testing

quevon24:
## Summary

Three small changes that let a developer flip `RUNPOD_ENABLED=True` in `.env.dev` and exercise the full pipeline end-to-end against a real RunPod endpoint, with the same S3 recovery behavior as production.

- `make_dev_data` now skips seeding when `RUNPOD_ENABLED=True` (placeholder PDFs would fail on the worker). The non-DEVELOPMENT guard also softens from `sys.exit(1)` to a silent `return`, so the `web-dev` entrypoint chain keeps starting the container cleanly even when seeding is intentionally a no-op (RunPod mode, or anywhere `DEVELOPMENT` is False).
- `s3_sync._s3_enabled()` now activates in `DEVELOPMENT=True` when `RUNPOD_ENABLED=True`, so processing artifacts round-trip to the dev bucket. Regular dev is unchanged.
- New "Testing RunPod Locally" section in `README.md` with required env vars, two AWS-credential options, tuning knobs, and a Quick Recipe.

### DIFF
--- a/README.md
+++ b/README.md
@@ -494,6 +494,110 @@ automatically.
 Custom component classes: `.btn-primary`, `.btn-outline`, `.btn-danger`,
 `.btn-ghost`, `.card`, `.input-text`, `.alert-*`, `.badge-*` (status badges).
 
+### Testing RunPod Locally
+
+The daemon can offload the two GPU-heavy blackletter steps (`detect` and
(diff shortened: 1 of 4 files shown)


grossir (approved):
LGTM!

PR #71, opened 2026-05-14
Tweak Javascript to approve detections

flooie:
Add small ui tweak to scanning portal

Also sync detections to disk every time.

### DIFF
--- a/scanning/services.py
+++ b/scanning/services.py
@@ -661,6 +661,76 @@ def _pull_processing_files_from_s3(scan_pk: int) -> None:
         )
 
 
+# Detection labels that map 1:1 to a redaction rect, keyed by fill color.
+# Mirrors blackletter's _REDACT_WHITE/_REDACT_BLACK sets; KEY_ICON is also
+# treated as black-filled here (blackletter has special "white before
+# caption" logic at generate time that we don't replicate at edit time).
+_SIMPLE_REDACT_RECTS: dict[str, str] = {
+    "KEY_ICON": "black",
+    "PAGE_HEADER": "white",
+    "STATE_ABBREVIATION": "white",
+    "DIVIDER": "black",
+    "HEADNOTE_BRACKET": "black",
+    "EDITORIAL": "black",
+}
+
+
+def _sync_redaction_rects_from_detections(scan_pk: int) -> None:
+    """Rebuild per-detection redaction rects from active Detection rows.
+
+    Only touches the 1:1 detection-to-rect entries (KEY_ICON, PAGE_HEADER,
+    STATE_ABBREVIATION, DIVIDER, HEADNOTE_BRACKET, EDITORIAL). Headnote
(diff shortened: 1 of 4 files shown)


scanning/static/scanning/viewer_step2.js
```diff
@@ -1773,6 +1773,44 @@ document.addEventListener('DOMContentLoaded', function () {
...
+                body: JSON.stringify({detection_id: det.id}),
+            }).then(function(r) {
+                if (!r.ok) throw new Error('approve failed: ' + r.status);
+                return r.json();
+            }).then(function(data) {
+                approveBtn.textContent = '✓';
+                det.confidence = 1.0;
+                var lbl = div.querySelector('.detection-label');
+                if (lbl) lbl.textContent = det.label + ' 1.0';
+                var sidebarItems = document.querySelectorAll('[data-unmatched-page="' + det.page_index + '"][data-unmatched-label="' + det.label + '"]');
```
quevon24, line 1796: Greying by page+label affects all matching sidebar items, not just the approved one. If a page has two detections with the same label, both get greyed. Match by data-detection-id instead.

scanning/services.py
```diff
@@ -661,6 +661,76 @@ def _pull_processing_files_from_s3(scan_pk: int) -> None:
...
+                "x0": round(d.x0, 1),
+                "y0": round(d.y0, 1),
+                "x1": round(d.x1, 1),
+                "y1": round(d.y1, 1),
+                "fill": _SIMPLE_REDACT_RECTS[d.label],
+                "type": d.label,
+            }
+        )
+
+    rects = [e for e in rects if e["rects"]]
```
quevon24, line 730: New page entries are appended at the end, so redaction_rects can end up out of page_index order after a delete-then-add. Maybe add a final sort to keep diffs clean.

scanning/views_api.py
```diff
@@ -767,64 +774,69 @@ def add_single_detection(request: HttpRequest, pk: int) -> JsonResponse:
...
+            active=True,
+            page_index=det["page_index"],
+            label_id=det["label_id"],
+            x0__gte=det["bbox"][0] - 15,
+            x0__lte=det["bbox"][0] + 15,
+            y0__gte=det["bbox"][1] - 15,
+            y0__lte=det["bbox"][1] + 15,
+        )
+        .order_by("pk")
+        .first()
```
quevon24, line 805: The old code updated all matching rows in the bbox window. The new code boosts only the first one (`.order_by("pk").first()`). The old behavior looks like it was guarding against duplicate active detections in the same area (e.g. two pipeline runs both leaving an active row). Is that scenario possible here? If yes, we shoul `.update(confidence=1.0)` on the queryset instead of .first() so we dont leave a stale low confidence twin behind. If duplicates can't happen, the new code is fine as-is, just worth confirming.

PR #72, opened 2026-05-15
feat(scanning): Bump Blackletter and remove masked directory

flooie:
Add some tests.
Remove masked output and directory.  This touches tests and a few places in javascript and python.  Add new blackletter feature to generate the LLM files as well.

Also remove done->list button which was not useful.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ dependencies = [
   "gunicorn",
   "pillow",
   "psycopg[binary,pool]",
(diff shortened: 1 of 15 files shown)


quevon24 (approved):
LGTM. The PR does what it sets out to do: bumps blackletter to 0.0.10 (turns on llm=True for per-page splits) and removes the masked PDF stuff from forms, views, JS, templates, and tests.

flooie:
thank you

PR #75, opened 2026-05-20
fix(scanning): close race between generate files and viewer reload

### LINKED ISSUE #74: Step 3: PDFs can 404 briefly after clicking Generate Files
After clicking Generate Files on step 3, the iframe can show "Error loading PDF: Missing PDF ..." for a few seconds before the file actually loads. Switching to another opinion (or waiting) usually resolves it without any user action.

## Root cause

In `services.py:run_generate_files`, the daemon currently does the work in this order:

1. Update `scan.opinions_json` and `redacted_pdf_path`.
2. Flip `scan.status` to `PENDING_REVIEW` and save.
3. Delete and recreate `OpinionScan` rows.
4. Push processing files to S3.

The frontend's progress poller (`viewer_progress.js`) reloads the page as soon as it sees `status == pending_review`. With the current order, the reload can land while:

- `OpinionScan` rows are mid-recreation (deleted but not yet replaced), or
- `OpinionScan` rows exist with valid `field.name` but the file is not yet on S3, so `_resolve_opinion_pdf` returns `None` and the iframe gets a 404.

## Proposed fix

Two layers:

1. Move the status flip in `run_generate_files` to be the very last step, after both the `OpinionScan` rows are created and `_push_processing_files_to_s3` has completed. This means the polling reload only fires when everything is ready.
2. In `viewer_step2.js`, retry `pdfjsLib.getDocument()` up to 3 times with 1s/2s/3s backoff when the error is `MissingPDFException`, as a safety net for any residual race (S3 eventual consistency, brief storage cache). The UI shows "Loading PDF (retrying N/3)..." between attempts.

The reorder alone closes most of the window; the retry covers the rest.

## Repro

1. Open a scan on step 3 with detections already in place.
2. Click "Generate Files" (or "Regenerate Files" if files already exist).
3. When the page reloads to step 3, click any opinion in the sidebar immediately.
4. Observe "Error loading PDF: Missing PDF ..." for a few seconds, then the file loads after switching opinions or waiting.

quevon24:
Closes #74.

Two-layer fix for the "Missing PDF" flicker on step 3 right after Generate Files:

- `services.py`: move the `PENDING_REVIEW` status flip to be the last step in `run_generate_files`, after `OpinionScan` rows are recreated and S3 push completes. The poll-and-reload now lands on a fully-ready state.
- `viewer_step2.js`: retry `pdfjsLib.getDocument()` up to 3 times with 1s/2s/3s backoff on `MissingPDFException` as a safety net for any residual race.

### DIFF
--- a/scanning/services.py
+++ b/scanning/services.py
@@ -2166,15 +2166,8 @@ def run_generate_files(scan_pk: int) -> None:
 
         scan.redacted_pdf_path = str(full_redacted) if full_redacted else ""
         scan.opinions_json = existing_opinions
-        # Files have been generated; the scan is ready for human review
-        # at step 3. The human clicks Approve to flip status -> APPROVED.
-        scan.stage = Stage.APPROVED
-        scan.status = Status.PENDING_REVIEW
-        scan.s3_uploaded = False
-        scan.progress_message = f"Generated {opinion_count} opinions"
-        scan.progress_log = ""
+        scan.progress_message = "Saving opinion records..."
         scan.save()
-        refresh_volume_queue_status_for_scan(scan)
 
         OpinionScan.objects.filter(scan=scan).delete()
         for i, op in enumerate(existing_opinions):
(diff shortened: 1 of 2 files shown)


grossir (approved):
LGTM!

PR #78, opened 2026-05-28
perf(admin): improve scan deletion and speed up change view

quevon24:
While testing in production I tried to delete a scan from the admin and the site stopped loading. I also noticed the scan change page (`/admin/scanning/scan/<id>/change/`) had gotten slow to load. Digging into both pointed at the same root cause: a processed `Scan` carries a lot of heavy data, and `ScanAdmin` was loading all of it eagerly.

## Slow change form

The change form had no `fieldsets`/custom form, so it rendered every field as editable, including the large per-page JSON/text payloads on `Scan`. On a processed scan these are megabytes each, so the server has to serialize and HTML-escape them into `<textarea>`s and the browser then has to render them.

Fields now excluded from the form (still stored in the DB and untouched by the pipeline):

- `ocr_results`
- `opinions_json`
- `page_map`
- `missing_pages`
- `margin_rects`
- `redaction_rects`
- `process_output`

Separately, `volume_obj` was rendered as a full `<select>` that loads every `Volume`; moved it into `raw_id_fields` to match `reporter`/`uploaded_by`.

## Delete takes the site down

Deleting a scan (single or the bulk "Delete selected scans" action) first renders the admin delete confirmation page. That page uses Django's collector to instantiate every related object so it can show the full "these will also be deleted" tree. A `Scan` cascades to `Page`, `Detection`, `Issue`, `PageInsert` and `PageDeletion`, and `Detection` alone can be tens of thousands of rows per scan. Collecting and rendering all of them exhausts memory and takes the worker down, before you even confirm.

Fix: override `get_deleted_objects()` to summarize the cascade as per-model counts instead of listing every row. Clickable links are kept only for the bounded sets a human would actually click (the selected scans, plus any blocking `OpinionScan` rows). Since `OpinionScan.scan` is `PROTECT`, a scan with opinions is now reported as protected on the confirmation page instead of 500-ing on the real delete.

The confirmation page now shows a "Summary" of counts (e.g. `Detections: 259`) and lists only the selected scan(s) under "Objects".

### DIFF
--- a/scanning/admin.py
+++ b/scanning/admin.py
@@ -1,6 +1,8 @@
 from django.contrib import admin, messages
-from django.urls import reverse
+from django.contrib.admin.utils import quote
+from django.urls import NoReverseMatch, reverse
 from django.utils.html import format_html
+from django.utils.text import capfirst
 
 from scanning.models import (
     Detection,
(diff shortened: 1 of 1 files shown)


scanning/admin.py
```diff
@@ -74,16 +76,99 @@ class ScanAdmin(admin.ModelAdmin):
...
         "date_created",
         "date_modified",
         "processed_at",
         "pages_link",
     ]
+    # Per-page payload fields hold megabytes on a processed scan.
+    # Rendering them as editable textareas makes the change view crawl,
+    # so keep them off the form (they stay in the DB and are used by the
+    # pipeline).
+    exclude = [
```
grossir, line 90: I think you should put process_output and missing_pages in `readonly_fields` so they remain inspectable. 

Exclude will make them not show

grossir (approved):
Looks good!

PR #79, opened 2026-05-28
fix(scanning): cap web pod memory growth via gunicorn recycling and streamed PDF exports

grossir:
Address some memory problems

- Recycle gunicorn workers more aggressively in docker/django/docker-entrypoint.sh: lower default MAX_REQUESTS 2500 → 500 and add --preload. At the observed ~200 MB/min growth in prod, 2500 was effectively inactive because workers OOMed before recycling; 500 fires the recycle while the leaked footprint is still under the per-worker budget. --preload keeps recycle cost low by COW-sharing heavy imports (fitz, cv2, blackletter) from the gunicorn master. Both are env-overridable.

- Stream export_pdf from disk instead of buffering in RAM (scanning/views_api.py). The view now saves the corrected PDF to a NamedTemporaryFile and serves it via StreamingHttpResponse with a generator that unlinks the temp file after the response finishes (and on the error path). Previously a 500-page corrected volume kept 50-200 MB of bytes in the worker's heap until the response flushed — slow clients amplified this under UvicornWorker.

- Release PyMuPDF pixmap immediately in serve_original_crop (scanning/views_process.py): set pix = None after tobytes() so the C-side image buffer (5-20 MB per crop at dpi=300) is freed at end-of-scope rather than waiting on GC. The viewer hits this endpoint many times per page.

### DIFF
--- a/docker/django/docker-entrypoint.sh
+++ b/docker/django/docker-entrypoint.sh
@@ -32,9 +32,10 @@ case "$1" in
         --group www-data \
         --workers ${NUM_WORKERS:-4} \
         --worker-class scanning.workers.UvicornWorker \
+        --preload \
         --no-control-socket \
         --timeout 180 \
-        --max-requests ${MAX_REQUESTS:-2500} \
+        --max-requests ${MAX_REQUESTS:-500} \
         --max-requests-jitter 100 \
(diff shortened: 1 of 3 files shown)


quevon24 (approved):
LGTM, nice improvements

PR #81, opened 2026-05-29
fix(scanning): guard pending page changes and refresh action bar in place

quevon24:
## Summary

After deleting a page in the process viewer, the action bar still showed the stale "Recheck" / "Re-validate" / "Next: Detect" buttons until a manual reload. None of those actions apply pending page changes (only "Rebuild & Validate" does), so clicking them silently stranded the deletion, and "Re-validate" would spend a full RunPod run while ignoring it. This adds a server-side guard and refreshes the action bar in place.

## Context

A pending page change is recorded the instant a page is marked for deletion (a `PageDeletion` row), but only `run_reprocess` ("Rebuild & Validate") ever applies it. To clear up a related report: page **deletions never trigger RunPod** (they run locally). RunPod is only invoked for page **insertions**, which OCR/validate each inserted page on the GPU. Pure deletions going back through the daemon queue is the local, no-cost path.

## Changes

- Block validate/detect/recheck when page changes are pending, redirecting to step 1 with a warning.
- Extract the action bar into a partial and add a `process_actions` JSON endpoint.
- Refresh the action bar and Pending Changes badge/banner in place after delete/undo.
- Add tests for the guard and the fragment.

### DIFF
--- a/scanning/views_process.py
+++ b/scanning/views_process.py
@@ -6,6 +6,7 @@ import os
 from pathlib import Path
 
 import fitz
+from django.contrib import messages
 from django.contrib.auth.decorators import login_required
 from django.http import (
     FileResponse,
@@ -15,6 +16,7 @@ from django.http import (
     JsonResponse,
 )
 from django.shortcuts import get_object_or_404, redirect, render
+from django.template.loader import render_to_string
 from django.urls import reverse
 from django.views.decorators.http import require_POST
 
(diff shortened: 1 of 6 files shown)


scanning/views_process.py
```diff
@@ -489,6 +491,82 @@ def serve_original_crop(request: HttpRequest, pk: int) -> HttpResponse:
...
+    scan = get_object_or_404(Scan, pk=pk)
+    try:
+        step = int(request.GET.get("step", 1))
+    except ValueError:
+        step = 1
+    if step < 1 or step > 3:
+        step = 1
+
+    has_pending_inserts = scan.inserts.exists()
+    has_pending_changes = scan.deletions.exists() or has_pending_inserts
```
grossir, line 550: these + has_detections is the same as in `scan_process_view`. It would be nice to abstract it out eventually

https://github.com/freelawproject/scanning/blob/8539f3e17bb3407d56c878df44649bf3fb81eb03/scanning/views_process.py#L189-L191

grossir (approved):
LGTM

PR #84, opened 2026-05-29
fix(runpod): make presigned-URL PDF downloads resumable and validate before processing

### LINKED ISSUE #82: RunPod handler: Add retries and partial download handling
We saw at least one big file failing download on RunPod. We expect more big files, so let's make the download of the pre signed URL more robust

Sentry Issue: [SCANNING-1X](https://freelawproject.sentry.io/issues/7512776236/?referrer=github_integration)

```
IncompleteRead: IncompleteRead(570425344 bytes read, 1348712520 more expected)
  File "urllib3/response.py", line 903, in _error_catcher
    yield
  File "urllib3/response.py", line 1049, in _raw_read
    raise IncompleteRead(self._fp_bytes_read, self.length_remaining)

ProtocolError: ('Connection broken: IncompleteRead(570425344 bytes read, 1348712520 more expected)', IncompleteRead(570425344 bytes read, 1348712520 more expected))
(1 additional frame(s) were not displayed)
...
  File "urllib3/response.py", line 1257, in stream
    data = self.read(amt=amt, decode_content=decode_content)
  File "urllib3/response.py", line 1112, in read
    data = self._raw_read(amt)
  File "urllib3/response.py", line 1027, in _raw_read
    with self._error_catcher():
  File "contextlib.py", line 158, in __exit__
    self.gen.throw(value)
  File "urllib3/response.py", line 927, in _error_catcher
    raise ProtocolError(arg, e) from e

ChunkedEncodingError: ('Connection broken: IncompleteRead(570425344 bytes read, 1348712520 more expected)', IncompleteRead(570425344 bytes read, 1348712520 more expected))
  File "handler.py", line 522, in handler
    result = fn(inputs, tmp_dir)
  File "handler.py", line 416, in _action_analyze
    _download_pdf(pdf_url, pdf_path)
  File "handler.py", line 264, in _download_pdf
    for chunk in r.iter_content(chunk_size=1024 * 1024):
```

grossir:
Solves #82

Large scan PDFs (multi-GB, 1000+ pages) take long enough that a single streamed download may die mid-transfer, failing the whole RunPod job. This makes the download
resilient and adds a structural integrity check before the file is handed to detect/analyze.

Resumable download (_download_pdf)
- Retry loop (default 5 attempts) with exponential backoff + jitter; resumes via Range: bytes=<offset>- instead of restarting.
- Handles all Range outcomes: 206 → append, 200-on-resume → server ignored Range, truncate and restart from 0 (prevents file corruption), 416 → already complete, 403 →
expired URL, fail fast so the daemon re-queues with a freshly-signed URL.
- Split connect/read timeouts so a stalled socket raises promptly and triggers a resume instead of hanging.
- Captures the expected total (Content-Length, or Content-Range total on 206) and raises if the file is short after the loop — a truncated PDF is never processed.

PDF validation (_validate_pdf, defense-in-depth)
- Replaces _page_count; called after download in both actions.
- Cheapest-first checks: non-empty → %PDF- header → %%EOF trailer in the last 2 KB (the truncation signal) → fitz.open with page_count >= 1.
- Guards the silent-corruption case where PyMuPDF opens a truncated PDF via a rebuilt xref and reports far fewer pages than the real document.
- On failure raises ValueError, which marks the job FAILED with no error_code so the daemon re-queues with a fresh URL.

Tests
- New scanning/tests/test_runpod_handler.py (16 tests) covering resume-after-drop, server-ignores-Range, 416-complete, 403 fast-fail, incomplete-body, give-up-after-max,
and PDF validation (valid single/multi-page, truncated, empty, non-PDF).
- Loads the standalone handler with its GPU-only imports stubbed; runs in the docker compose env. Full scanning.tests suite (222 tests) passes.

Notes
- No blackletter change needed: the presigned TTL is already 86400s, and a raised handler exception already maps to the daemon's transient/re-queue path. Validation is
scoped to the handler (where the download-truncation risk lives); blackletter's local-mode callers read a locally-produced bitonal.pdf.
- Still pending (not in this PR): manual end-to-end rerun of scan 1835 (~1.8 GB / 1307 pages) on a real GPU worker.

### DIFF
--- a/scanning/runpod/handler.py
+++ b/scanning/runpod/handler.py
@@ -425,17 +425,83 @@ def _download_pdf(url: str, dest: Path) -> None:
             )
 
 
-def _page_count(pdf_path: Path) -> int:
-    """Return the number of pages in a PDF.
+# Trailer window scanned for the %%EOF marker. A conforming PDF ends with
+# %%EOF (optionally followed by whitespace); incrementally-updated files
+# carry several, and only the last one matters. 2 KB comfortably covers
+# trailing newlines and a final cross-reference stream.
+_PDF_EOF_WINDOW = 2048
 
-    :param pdf_path: Path to the PDF on disk.
+
+def _validate_pdf(pdf_path: Path) -> int:
+    """Validate that ``pdf_path`` is a complete, openable PDF and return
+    its page count.
+
+    Defense-in-depth against a truncated download slipping through: the
+    resumable ``_download_pdf`` already checks the byte count against the
+    server's advertised size, but that only fires when the server sends a
+    size. The real hazard is a partial PDF that PyMuPDF can still *open*
+    by rebuilding a broken xref, reporting far fewer pages than the real
+    document — which would then silently produce wrong detections / page
+    analysis. Checking the header and the ``%%EOF`` trailer turns that
+    silent corruption into a hard failure.
+
+    A raised error propagates out of the action and (via ``handler()``)
+    marks the RunPod job FAILED with no ``error_code``; the daemon treats
+    a bare FAILED as transient and re-queues the scan with a freshly
+    signed URL. So a truncated download is re-fetched automatically; a
+    genuinely corrupt source PDF loops until the daemon's retry ceiling
+    and then surfaces as an error.
+
+    Checks run cheapest-first so a bad file fails before the fitz open.
+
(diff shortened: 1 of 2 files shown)


quevon24 (approved):
Looks great. This will really help with the big multi-GB scans that kept dying mid-download

PR #85, opened 2026-05-30
fix(scanning): serve processed PDF in viewer and persist bitonal/OCR to S3

quevon24:
## Problem

The scan viewer (`serve_scan_pdf`) was streaming the full original PDF even when a much smaller processed copy existed. For multi-GB / 1000+ page scans that is an order-of-magnitude larger download than necessary (e.g. 20 MB original vs ~2.4 MB bitonal).

Two root causes:

1. `serve_scan_pdf` already intended to prefer the OCR/bitonal PDF, but it returned the original as soon as a local copy was resolvable, short-circuiting before the S3 lazy-pull that fetches the processed PDF. So whenever the original was on disk but the bitonal/OCR weren't co-located, it served the big original.
2. The bitonal and OCR PDFs were only uploaded to S3 by the end-of-pipeline push. A crash, or a `reprocess` run (which never does the full push), left them out of S3 entirely, so there was nothing for the viewer to pull.

## Changes

`serve_scan_pdf` now prefers the processed PDF and only falls back to the original after attempting the pull: serve OCR/bitonal if local, otherwise pull processing files from S3 and re-check, and only then fall back to the original. The big original can no longer pre-empt a fetchable bitonal.

`_ensure_bitonal` and `_run_ocr` now upload their output to S3 the moment it is generated, via a new `_push_generated_file_to_s3` helper (same credential guard as `_push_processing_files_to_s3`). This makes the derived files resilient to a later crash and available to any reprocess/viewer pull. In local dev the upload is a no-op and the file is already on the shared media mount, so it is served locally without S3.

Together this closes the loop: generate → push immediately, and re-entry/viewer → pull if missing → serve the small file in preference to the original.

## Notes

This applies to future generations. Scans whose bitonal/OCR were never produced by a prod-mode run still need a full Re-validate to populate S3 before the viewer can serve the small file; the pull only fetches what is actually in S3.

### DIFF
--- a/scanning/services.py
+++ b/scanning/services.py
@@ -291,6 +291,9 @@ def _ensure_bitonal(scan: "Scan", output_dir: Path) -> Path:
             str(output_dir),
             progress_callback=_bitonal_progress,
         )
+        # Persist as soon as it exists so a later crash, or a reprocess
+        # that skips the end-of-pipeline push, still leaves it in S3.
+        _push_generated_file_to_s3(scan.pk, "bitonal.pdf")
 
     with fitz.open(str(bitonal_path)) as pdf:
         page_count = pdf.page_count
@@ -324,7 +327,7 @@ def _run_ocr(
         current=0,
         total=total_pages,
     )
-    return _ocr(
+    ocr_path = _ocr(
         scan_pk,
         bitonal_path,
         output_dir,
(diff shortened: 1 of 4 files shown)


grossir (approved):
LGTM!

PR #92, opened 2026-06-09
feat(scanning): queue stats, retire scan detail view, volumes-only import

### LINKED ISSUE #91: Mismatch between Volume and Scan metadata?
See these numbers on the queue page https://scanning.courtlistener.com/queue/ which reflect Volume status
Image

And the numbers from the Scan table on #38 , both from today. They should be roughly the same? 

```sql
scanning=> select status, count(*) from scanning_scan group by status;
     status     | count 
----------------+-------
 pending_review |    90
 processing     |     1
 queued         |     1
(3 rows)
```

quevon24:
Groups several related queue/UI improvements that came out of reviewing #91.

- **Show `scanned` and `unavailable` in the queue stats (#91).** The queue header only displayed `needs_scanning / assigned / scanning / complete`, so volumes in `scanned` or `unavailable` were counted in the total but absent from the breakdown, which is what made the queue numbers look off. Both buckets are now in the stats line and the filter tabs. `scanned` is the "scanning done, awaiting review" state; volumes leave it for `complete` as their scans get approved in the viewer.
- **Retire the standalone scan detail page.** `/scans/<pk>/` showed metadata and a review form but no PDF, and already redirected processed scans away. It now redirects to the process view (which shows the PDF and handles review/approval). The URL name is kept so old/bookmarked links still work.
- **Repoint links to the process view.** The list "View" button, mobile card, opinion detail link, and pages breadcrumb now go straight to the process view; the dead detail template and its unused review-form handling were removed.
- **Seed volumes only in `import_scanlist`.** The command no longer creates placeholder `Scan` rows (status `uploaded`, no file). It populates `Volume` records only.
- **Skip existing volumes on import.** A volume already in the DB is skipped with a message rather than overwritten, so re-runs never clobber manual edits. Multi-part rows (parts A/B, advance sheets) within a single run still merge into one volume and widen its page range.
- **Remove dead upload/review code.** Dropped the unreachable `scan_upload` view (it had no URL route) along with its template and `ScanUploadForm`, plus the unused `ScanReviewForm` and `OpinionScanReviewForm`.
- **Isolate test media in a temp dir.** `ScanFactory`'s `original_pdf` is a real file field, and tests that didn't override `MEDIA_ROOT` were leaking dummy PDFs into `scanning/assets/media/`. Tests now use a temp `MEDIA_ROOT`, and `original_scans/` / `page_inserts/` are gitignored.
- **Paginate the queue view.** The queue listed every matching volume and prefetched all their scans in one request, which grows with each upload. It now paginates at 100 volumes per page. The stats line still aggregates over all volumes, so totals stay accurate.

Closes #91


grossir (approved):
LGTM!

PR #93, opened 2026-06-10
feat(scanning): show upload progress on queue uploads

### LINKED ISSUE #86: Give uploaders some feedback on file upload progress
Usually scans are big files, in the range of 1 to 2 GB. Uploaders don't get much feedback, the page just hangs while uploading. Some clearer user feeback would be nice

quevon24:
Fixes #86

Scan uploads are 1-2 GB PDFs submitted through a plain form POST, so the page just hung with no feedback for the several minutes the transfer takes.

The upload form on the queue detail page now submits via `XMLHttpRequest` and renders a progress bar with a live percentage from the browser's upload progress events. Once the file finishes sending, the label switches to "Processing upload, this may take a while…" to cover the server-side phase (writing the file to disk, and pushing it to S3 in prod), which previously looked like a freeze at 100%.

Details:

- `queue_upload` returns `{"redirect": url}` as JSON when the request comes from the XHR (detected via `X-Requested-With`), and the script navigates with `window.location.assign()`. If the XHR followed the normal 302 itself, Django's success/error messages would be consumed by the throwaway XHR response and never shown to the user. Non-JS form posts still get the regular redirect, so the form keeps working without JavaScript.
- The clicked submit button's `action` value is appended to the `FormData` manually, so Upload & Validate vs Upload Only still works (with a fallback for browsers without `SubmitEvent.submitter`).
- Submit buttons are disabled while an upload is in flight, and a `beforeunload` guard warns before leaving the page mid-upload, since refreshing aborts the request and silently loses the upload.
- Network and HTTP failures show an inline error and re-enable the form.

Verified locally against the dev server with 1 GB and 2.2 GB files (the latter to rule out 2 GiB content-length issues); both completed and returned the JSON redirect, and queued messages render on the page after navigation.

Captura desde 2026-06-10 17-04-06

Captura desde 2026-06-10 17-04-10

### DIFF
--- a/scanning/templates/scanning/queue_detail.html
+++ b/scanning/templates/scanning/queue_detail.html
@@ -82,7 +82,7 @@
 <!-- Upload form -->
 <div class="card mb-6">
   <h2 class="text-sm font-bold mb-3">Upload a scan</h2>
-  <form method="post" action="{% url 'queue_upload' reporter_slug=volume.reporter.short_name vol=volume.volume_number %}" enctype="multipart/form-data">
+  <form id="upload-form" method="post" action="{% url 'queue_upload' reporter_slug=volume.reporter.short_name vol=volume.volume_number %}" enctype="multipart/form-data">
     {% csrf_token %}
     <input type="hidden" name="new_scan" value="1">
     <div class="grid grid-cols-2 sm:grid-cols-4 gap-2 mb-3">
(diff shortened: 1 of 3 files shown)


grossir (approved):
LGTM. 

I will open another issue to track the "Processing upload" part, and see if we can make that work better, since the user is stuck waiting for an action (S3 upload) that doesn't need them, after they already uploaded the file to the server

PR #95, opened 2026-06-12
fix(scanning): render step-2 overlays by pdf index

### LINKED ISSUE #90: Research bad detections / redactions on prod data
Will not post screenshots, but there are some obvious bad detections on the first few pages of this [scan](https://scanning.courtlistener.com/scans/1875/process/?step=2). There may be others to be flagged.

May be related to 
- it being a full uploads (having the title page, TOC, etc)
- the zoom level of the viewer? I noticed it changes a bit when zooming out / in

quevon24:
## What happened

On volumes that start with unnumbered front matter (cover, cases-reported lists), the step-2 review viewer drew the redaction, margin, and detection overlays on the wrong pages: boxes belonging to the first content pages were stacked on the front matter, and the content pages themselves showed nothing.

## Why it happened

Pages with no detected page number fall back to their PDF position as their logical number, so the front matter claims logical pages 1..N while the real book pages 1..N exist later in the PDF. The viewer keyed its page elements and overlay lookups by logical number, which produced duplicate DOM ids and ambiguous lookups that always resolved to the first (front matter) page. It also meant manual redactions drawn on one of the colliding pages could be saved with the other page's index.

This also explains the zoom observation in the issue: the initial overlay pass and the re-render path used after zooming resolved the ambiguity differently, so the boxes on the affected pages visibly changed when zooming in or out. Zoom was just the trigger that switched drawing paths, not a separate bug.

## What changed

All overlay rendering, navigation to detections, and save payloads in the step-2 viewer now resolve pages by their PDF index, which is unambiguous regardless of how logical numbers come out. Logical page numbers are still used, but only for display labels and sidebar links. Overlays were verified to keep the same counts and positions across zoom levels and scroll cycles after the change.

The stray detections that remain on front-matter pages (e.g. the publisher's key logo detected as a key icon) are model false positives in the data, not display bugs; they can be cleaned up with the existing delete/suppress tools.

Related to #90

### DIFF
--- a/scanning/static/scanning/viewer_sidebar.js
+++ b/scanning/static/scanning/viewer_sidebar.js
@@ -232,9 +232,11 @@ var _opinions = [];
                         r.bottom > viewerRect.top &&
                         r.top < viewerRect.bottom
                     ) {
-                        bestPageNum = parseInt(
-                            pages[i].dataset.pageNum || "0"
-                        );
+                        // data-caption-page / data-key-page on the cards
+                        // are 1-based PDF pages, so compare pdf index + 1
+                        // (logical numbers can repeat across the scan).
+                        bestPageNum =
+                            parseInt(pages[i].dataset.pdfIndex || "-1") + 1;
                         break;
                     }
                 }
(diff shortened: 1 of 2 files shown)


grossir (approved):
LGTM!

PR #96, opened 2026-06-12
feat(scanning): show loading spinner in opinion pdf viewer

quevon24:
Adds a loading state to the PDF viewer on the opinion detail page. When a PDF isn't cached locally on the server, the first request triggers a lazy S3 pull of the scan's entire processing directory, which can take a while. Until now the viewer just sat blank during that window.

- Show an animated spinner in the PDF container while PDF.js fetches the document, hidden once pages start rendering.
- Show an error message instead of spinning forever if the fetch fails.
- Add a load-token guard to the PDF.js callbacks. This fixes two pre-existing races: switching tabs during a slow load no longer interleaves pages from both variants into the container, and opinions with only an original PDF no longer render every page twice on initial load (it was loaded once on DOMContentLoaded and again via the tab switch).


grossir (approved):
LGTM

PR #97, opened 2026-06-15
fix(scanning): downgrade expected NO_GPU logs, error only on exhausted retries

### LINKED ISSUE #89: Research "GPU not available" errors
Sentry Issue: [SCANNING-1M](https://freelawproject.sentry.io/issues/7509879762/?referrer=github_integration)

```
GPU not available; skipping weight/model preload. Jobs will return error_code=NO_GPU; scans are re-queued automatically by the daemon.
```

quevon24:
The `NO_GPU` path is expected and fully self-healing: the worker fitness check (`_require_gpu`) keeps GPU-less workers out of the pool, and any job that slips through returns `error_code=NO_GPU`, which the daemon classifies as transient and auto re-queues (up to `RUNPOD_MAX_TRANSIENT_RETRIES`). Despite that, the worker logged it at `error` level, so it surfaced in Sentry (SCANNING-1M: 707 events over 18 days, 0 users impacted) as pure noise.

This shifts the log levels so Sentry only fires when something is actually actionable:

- `_preload` and the per-job reject in the handler now log at `warning` (expected and self-healing, breadcrumb only, no Sentry event).
- The daemon's `ERROR_MAX_RETRIES` branch now logs at `error`, so Sentry alerts exactly once when a scan exhausts its transient retries, which is the signal for a real RunPod capacity shortage worth investigating.

Verified against Sentry that no `ERROR_MAX_RETRIES` or per-job-reject issues exist today, confirming the self-healing works and this was log noise rather than a correctness problem.

Closes #89

### DIFF
--- a/scanning/runpod/handler.py
+++ b/scanning/runpod/handler.py
@@ -136,7 +136,12 @@ def _preload() -> None:
         # the scanning daemon classifies as transient and re-queues the
         # scan (up to RUNPOD_MAX_TRANSIENT_RETRIES attempts) for the next
         # tick to resubmit to a different worker.
-        logger.error(
+        #
+        # Logged at warning, not error: this is expected and fully
+        # self-healing, so it must not raise a Sentry event. The daemon
+        # escalates to error only if a scan actually exhausts its
+        # transient retries.
+        logger.warning(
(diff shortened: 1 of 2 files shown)


grossir (approved):
LGTM

PR #99, opened 2026-06-16
fix(scanning): avoid duplicate PDF fetch on opinion viewer load

### LINKED ISSUE #98: Original-only opinions fire a redundant PDF fetch on initial load
### Summary

On the opinion detail page, opinions that have only an original PDF (no redacted variant) issue the serve_opinionscan_pdf request twice on initial page load. Because that
endpoint lazily pulls the scan's entire processing directory from S3 on a cache miss, the duplicate request can compound load latency — the exact slowness the new loading
spinner (#96) is meant to cover.

### Where

scanning/templates/scanning/opinion_detail.html, the DOMContentLoaded handler:

document.addEventListener('DOMContentLoaded', function() {
    if (_currentUrl) loadPdf(_currentUrl);            // (1) loads original
    {% if not opinion.redacted_pdf and opinion.original_pdf %}
    var btn = document.querySelector('.pdf-tab');
    if (btn) switchTab(btn);                           // (2) calls loadPdf again
    {% endif %}
});

## Why it happens

For an original-only opinion, _currentUrl is already the original variant, so step (1) loads it. Step (2) exists only to apply the active-tab highlight to the Original tab
— the template hardcodes the active class on the Redacted tab only (lines ~17 vs ~21), so without it the lone Original tab renders as unselected. But switchTab also calls
loadPdf, so the same URL is fetched a second time.

The load-token guard added in #96 correctly prevents the duplicate render, but the second getDocument() network request still fires before any token check runs, so the
redundant S3-triggering fetch is not prevented.

### Suggested fix

Separate the two concerns in switchTab (loading vs. styling). Either:

- Guard the reload: if (url !== _currentUrl) loadPdf(url); inside switchTab, or
- In the {% if %} branch, set the active class on the button directly instead of calling switchTab, or
- Mark the active tab server-side based on the default variant, which removes the need for the DOM-load switchTab call entirely.

### Impact

Minor — wasteful duplicate request, most visible on cold S3 cache. No incorrect rendering (the token guard handles that). Surfaced during review of #96.

grossir:
## Summary

On the opinion detail page, opinions that have **only an original PDF** (no redacted variant) fetch the `serve_opinionscan_pdf` endpoint twice on initial load. Since that endpoint lazily pulls the scan's entire processing directory from S3 on a cache miss, the duplicate request compounds the load latency the new spinner (#96) is meant to cover.

## Why it happens

In `scanning/templates/scanning/opinion_detail.html`, the `DOMContentLoaded` handler loads `_currentUrl`, then — for original-only opinions — calls `switchTab` solely to apply the active-tab highlight (the template hardcodes the `active` class on the Redacted tab only). But `switchTab` also called `loadPdf`, firing a second fetch of the same URL.

The load-token guard from #96 suppresses the duplicate *render*, but the second `getDocument()` request still goes out before any token check runs.

## Fix

Guard the reload inside `switchTab` so it only fetches when the variant actually changed:

```js
if (url !== _currentUrl) loadPdf(url);
```

The styling update still runs unconditionally, so the active-tab highlight is unaffected. As a side benefit, re-clicking the already-active tab no longer triggers a redundant reload.

Closes #98.

### DIFF
--- a/scanning/templates/scanning/opinion_detail.html
+++ b/scanning/templates/scanning/opinion_detail.html
@@ -157,7 +157,11 @@ function loadPdf(url) {
 
 function switchTab(btn) {
     var url = btn.getAttribute('data-url');
-    loadPdf(url);
+    // Only (re)load when the variant actually changed. On initial load the
+    // DOMContentLoaded handler already loads _currentUrl, so the styling-only
+    // switchTab call for original-only opinions must not trigger a second
+    // fetch of the same PDF (the serve endpoint lazily pulls from S3).
+    if (url !== _currentUrl) loadPdf(url);
 
     document.querySelectorAll('.pdf-tab').forEach(function(t) {
         t.className = 'pdf-tab text-sm px-4 py-2 font-medium rounded-t border border-b-0 border-transparent text-gray-500 dark:text-gray-400 hover:text-gray-700 dark:hover:text-gray-200 hover:bg-gray-50 dark:hover:bg-gray-700/50 -mb-px';


quevon24 (approved):
LGTM

PR #101, opened 2026-06-17
fix(scanning): resolve viewer highlights/navigation by pdf_index and speed up step-2 jumps

### LINKED ISSUE #100: Process viewer: front-matter page-number collisions (steps 1–2) and slow step-2 page jumps
## Summary

The step 1 and step 2 process viewers misbehave when a scanned volume starts with unnumbered front matter, because logical page numbers are not unique: the unnumbered pages fall back to `logical = pdf_page` and borrow the numbers the real pages also carry, so anything keyed on the logical number lands on the wrong page. Separately, jumping to a page in step 2 is very slow. Related to #90.

## Example

Consider a volume whose first 13 PDF pages are unnumbered front matter (cover and "Cases Reported" tables), with opinions starting on PDF page 14. Document page 1 is PDF page 14, so PDF pages 14 to 26 are document pages 1 to 13.

## Bugs

**Front-matter collisions (logical page numbers not unique):**
1. False "DUPLICATE" badges on the real pages (step 1).
2. False red `flagged` borders on the real pages (steps 1 and 2).
3. The sidebar "Pages" list jumps to the wrong page (step 1).
4. Issue cards and opinion image/headnote badges jump to the wrong page (steps 1 and 2).

**Jump behavior / performance:**
5. Far jumps land off-target and need several clicks: lazy-rendered pages start at a placeholder height and shift as they render (both viewers; worsened by smooth scrolling).
6. Jumping to a page in step 2 takes ~4s: selecting an opinion redraws every redaction/margin box across the whole document, interleaving layout reads with DOM writes (a multi-second forced reflow).

## Root causes

- **Logical page numbers are not unique** → highlight and navigate by the unique `pdf_index`.
- **Lazy-render layout shift** → jump instantly, then re-align until the target settles (no smooth scroll).
- **Whole-document overlay redraw on opinion select** → those overlays are already drawn per page and don't change on selection; stop redrawing them on click, and draw margins per page on render.

quevon24:
## Summary

Fixes the step 1 and step 2 process viewers when a volume starts with unnumbered front matter (logical page numbers are not unique, so highlight/navigation keyed on them landed on the wrong page), and makes page jumps in step 2 fast. Related to #90.

## Changes

- Highlight flagged pages and navigate by `pdf_index` (not logical number) across both viewers: the sidebar "Pages" list, issue cards, and opinion image/headnote badges all resolve to the correct physical page (`data-flagged-pages` → `data-flagged-indices`).
- Jump without a scroll animation in both viewers: snap to the target, then re-align until it settles, instead of smooth-scrolling through every page in between.
- Fix the ~4s step-2 page jump: selecting/clearing an opinion no longer redraws every redaction/margin box across the whole document (that caused a ~1.4s forced reflow). Margins draw per page on render, and detection/margin lookups are O(1) per page.

## Notes

The false "DUPLICATE" badge from the same root cause is fixed in blackletter (freelawproject/blackletter#53); it takes effect once that is released and the dependency pin is bumped here.

Closes #100

### DIFF
--- a/scanning/static/scanning/shared.js
+++ b/scanning/static/scanning/shared.js
@@ -339,3 +339,20 @@ function drawExistingRedactions(overlay, pageRedactions, scale) {
         }
     });
 }
+
+// Jump to a lazy-rendered page element without a scroll animation. Pages start
+// at a placeholder height and grow to their true height once rasterized, which
+// shifts everything below them; so after jumping we re-align until the target
+// stops moving (capped) instead of animating through every page in between.
+window.scrollPageIntoView = function (el) {
+    if (!el) return;
+    var tries = 0;
+    (function settle() {
+        var before = el.getBoundingClientRect().top;
+        el.scrollIntoView({ block: 'start' });
+        var after = el.getBoundingClientRect().top;
+        if (++tries < 10 && Math.abs(after - before) > 2) {
+            setTimeout(settle, 90);
+        }
+    })();
+};
--- a/scanning/static/scanning/viewer_progress.js
+++ b/scanning/static/scanning/viewer_progress.js
@@ -80,9 +80,9 @@
             html +=
                 '<div class="rounded px-2 py-0.5 mb-0.5 cursor-pointer text-xs flex items-center justify-between ' +
                 bg +
-                "\" onclick=\"goToPage(" +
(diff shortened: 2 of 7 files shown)


grossir (approved):
LGTM

PR #104, opened 2026-06-23
fix(scanning): mark page_map duplicates in step-1 sidebar

### LINKED ISSUE #103: Step 1 sidebar page list and PDF viewer disagree on which pages are duplicates
### Summary
In the step 1 process viewer, duplicate pages flagged with a `DUPLICATE` badge in the PDF viewer are not marked in the left sidebar "Pages" list. The two views use different definitions of "duplicate", so they routinely disagree.

### Steps to reproduce
1. Open a scan whose `page_map` contains duplicate pages that are not adjacent in the numbering sequence (e.g. the same printed "page 1" appearing on several pages, common with front matter). Scan 1872 is an example.
2. Scroll the PDF viewer: several pages show the orange `DUPLICATE` badge.
3. Look at the sidebar "Pages" list: those same pages have no duplicate marker.

### Expected
The sidebar "Pages" list and the PDF viewer agree on which pages are duplicates, and the sidebar lets you spot and jump to them without scrolling the viewer.

### Actual
The sidebar shows no duplicate indicator for these pages. The only way to discover them is by manually scrolling the PDF viewer.

### Root cause
There are two independent duplicate signals on step 1:

- **PDF viewer** badge comes from `page_map.duplicate`: a detected page number that already appeared on an earlier page (anywhere in the document).
- **Sidebar "Pages" list** came from `seq_issue`, computed in `views_process.py`, which only flagged `diff == 0` between *consecutive* detected page numbers.

When duplicate copies are not consecutive (the usual case for repeated low numbers from unnumbered front matter, see #90), the consecutive-only check never fires, so the sidebar stays silent while the viewer shows a badge.

### Related context (not part of this bug)
While investigating, the false duplicate badges originally reported on scan 1872 (pages 19, 22, 23, 24, 38) turned out to be **stale `page_map` data** generated before the blackletter front-matter fix (v0.0.13, pulled in via #102). Reprocessing the scan regenerates a correct `page_map` (24 → 7 duplicate flags, all genuine). No code change is needed for that part; this issue is strictly about the sidebar/viewer inconsistency that remains even with correct data.

quevon24:
The step-1 sidebar "Pages" list and the PDF viewer disagreed on which pages are duplicates. The viewer badges pages from `page_map.duplicate` (a detected page number that appeared earlier anywhere in the document), while the sidebar derived its marker from `seq_issue`, a consecutive-only check (`diff == 0` between adjacent detected numbers). When duplicate copies are not adjacent (the common case for repeated low numbers borrowed from unnumbered front matter, see #90), the sidebar showed nothing while the viewer showed a badge, so the only way to find those pages was to scroll the viewer.

This drives the sidebar duplicate marker from the same `page_map` data the viewer uses: the view computes a per-page `is_duplicate` flag from the page_map, and `seq_issue` is now limited to ordering anomalies (backward / gap). The misleading consecutive "DUP" divider is replaced with an inline `DUP` badge plus the existing orange row highlight, so the sidebar matches the viewer. Clicking a duplicate row still jumps to the page and highlights it.

The live-processing preview (`viewer_progress.js`) intentionally keeps the consecutive heuristic, since the `page_map` does not exist yet at poll time and a naive "seen-before" check there would reintroduce the front-matter false positives.

Verified on scan 1872: the sidebar now shows 7 `DUP` badges matching the viewer (previously 0). Added `TestSidebarDuplicateMarkers` covering a non-adjacent page_map duplicate; full view suite passes.

Closes #103

### DIFF
--- a/scanning/templates/scanning/scan_process.html
+++ b/scanning/templates/scanning/scan_process.html
@@ -129,27 +129,22 @@
           <span class="text-[9px] text-red-600 dark:text-red-400 font-bold whitespace-nowrap">ORDER</span>
           <div class="flex-1 border-t-2 border-dashed border-red-400"></div>
         </div>
-        {% elif r.seq_issue == 'duplicate' %}
-        <div class="flex items-center gap-1 my-1">
-          <div class="flex-1 border-t-2 border-dashed border-orange-400"></div>
-          <span class="text-[9px] text-orange-600 dark:text-orange-400 font-bold whitespace-nowrap">DUP</span>
-          <div class="flex-1 border-t-2 border-dashed border-orange-400"></div>
-        </div>
         {% endif %}
       <div class="rounded px-2 py-0.5 mb-0.5 cursor-pointer text-xs flex items-center justify-between
            {% if not r.detected %}bg-red-50 dark:bg-red-900/20 border border-red-200 dark:border-red-800
            {% elif r.seq_issue == 'backward' %}bg-red-100 dark:bg-red-900/30 border border-red-300 dark:border-red-700
-           {% elif r.seq_issue == 'duplicate' %}bg-orange-100 dark:bg-orange-900/30 border border-orange-300 dark:border-orange-700
+           {% elif r.is_duplicate %}bg-orange-100 dark:bg-orange-900/30 border border-orange-300 dark:border-orange-700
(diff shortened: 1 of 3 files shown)


grossir (approved):
LGTM!

PR #106, opened 2026-06-23
fix(scanning): validate, clear, and refresh manual page-number edits

### LINKED ISSUE #105: Manual page-number correction (assign_page): can't clear a number, accepts invalid values, stale analysis
### Summary
The step-1 manual page-number correction (click a page's number → `prompt()` → `assign_page`) has several gaps: it can only *set* a number, never clear one; it accepts invalid values like `0`; and it doesn't refresh duplicate/sequence analysis or the `page_map`, so the QC view goes stale until a full Recheck.

### Where
- UI prompt handler: `scanning/static/scanning/viewer_step1.js:336-357`
- Endpoint: `scanning/views_process.py:761` (`assign_page`)

### Gaps

**1. No way to clear a page number (mark a page as unnumbered / front matter).** The prompt only proceeds on non-empty input (`if (num !== null && num.trim())`, `viewer_step1.js:339`), and `assign_page` only ever sets `detected`. A page OCR wrongly tagged with a number (e.g. a table-of-contents / "cases reported" page with no printed page number) can't be corrected to "no number", only re-pointed to a different one. The spurious number and any duplicate warning it triggers can't be removed without deleting the page (which would wrongly drop real content).

**2. No validation, accepts 0 / negatives / non-numeric.** `assign_page` stores `page_number` verbatim into `detected` (`views_process.py:779`) with no validation; the frontend only checks for non-empty. So `0`, negatives, and non-numeric strings all pass. Page numbers should be positive integers.

**3. Manual edits don't refresh duplicate/sequence analysis or the page_map.** `assign_page` mutates `ocr_results` and deletes the page's `NO_PAGE_NUMBER` issue, but doesn't rebuild `page_map` or recompute duplicate/sequence issues (`views_process.py:776-789`). Since the DUP badge/sidebar markers come from `page_map.duplicate` (#103), an edit that creates or resolves a duplicate isn't reflected until the user runs Recheck (`recalculate`). On reload the page shows `(manual)` but stale duplicate/sequence state.

**4. (minor) First occurrence of a repeated number isn't badged, inconsistent with the duplicate_page issue.** blackletter flags only the 2nd+ occurrence as `duplicate`, so in a group of three pages numbered "1" only two get a DUP badge, while the `duplicate_page` issue card lists all three (e.g. `[3, 9, 18]`). The page-level badge and the issue's page list disagree about the first occurrence.

### Downstream impact (verified)
The spurious number is largely a cosmetic QC warning, not an output-corrupting bug:
- Opinion pairing groups pages by caption→key detection position (`page_index`), not page number, so a duplicate/spurious number doesn't merge or split opinions. Confirmed on scan 1872: opinions start at the real page 1 (pdf page 18), and the front-matter "1"s on pdf pages 3/9 are not opinion boundaries.
- Detected numbers feed opinion page-range labels/filenames via `detections.json` → `_opinion_page_bounds` (`blackletter/scanner.py:1377-1389`), but only mislabel an opinion when the wrong number is on that opinion's caption or key page (front-matter pages are not).
- Manual edits reach generated output only when pairing/redaction is re-run; they don't retroactively update already-generated `opinions_json`.

quevon24:
The step-1 manual page-number correction (`assign_page`) had three gaps (#105): it could only *set* a number (never clear one), accepted invalid values like `0`, and didn't refresh the page_map, so duplicate flags went stale after an edit.

**Clear a number.** The edit prompt now accepts a blank value to mark a page as having no number (e.g. front matter the model mis-tagged): the page reverts to the "no page #" state and is recorded as `manual`. Clearing is explicit, sent as `null`/`""`.

**Validation.** `assign_page` now rejects non-positive and non-numeric values with `400`, unknown PDF pages with `404`, and missing `pdf_page`/`page_number` keys with `400` (so a malformed request can't silently clear a page). The prompt handler pre-validates and surfaces server errors.

**Refresh duplicate flags.** Each set/clear now rebuilds `page_map` and `missing_pages` from the current `ocr_results` via the new `services.rebuild_page_map`, so the viewer's DUPLICATE badge reflects the edit immediately instead of staying stale until a full Recheck. The rebuild is PDF-free (uses the same expected-range derivation as Recheck) and deliberately does not touch Issue rows, so dismissed issues are preserved; issue cards still refresh on a full Recheck. Extracted `_split_detected` and `_build_analysis` as shared helpers with `recalculate_issues` (no behavior change to the Recheck path).

Verified against the running app: validation returns the right 400/404s, a blank clears the number, and creating/removing a duplicate flips the rendered page_map's `duplicate` flag for the affected page.

Deferred to a separate blackletter issue: only the 2nd+ copies of a repeated number are flagged in the page_map while the `duplicate_page` issue lists all copies (the first-occurrence inconsistency, gap 4 in #105).

Closes #105

### DIFF
--- a/scanning/services.py
+++ b/scanning/services.py
@@ -1405,6 +1405,126 @@ def run_incremental_validation(scan_pk: int, pdf_path: str) -> None:
 # ---------------------------------------------------------------------------
 
 
+def _split_detected(
+    ocr_results: list, exp_start: int | None, exp_end: int | None
+) -> tuple[list, dict]:
+    """Partition detected single page numbers into out-of-range and in-range.
+
+    A detected number is out of range when it is below 1 or, when an expected
+    range is known, falls more than 5 outside it. Range pages and pages
+    without a detected number are skipped.
+
+    :param ocr_results: Per-page OCR results.
+    :param exp_start: Expected first page number, or None.
+    :param exp_end: Expected last page number, or None.
+    :returns: ``(out_of_range, seen_nums)`` where ``seen_nums`` maps each
+        in-range number to the list of PDF pages it appears on.
+    :rtype: tuple[list, dict]
+    """
+    out_of_range: list = []
+    seen_nums: dict = {}
+    for r in ocr_results:
+        if not r["detected"] or r.get("type") == "range":
+            continue
+        try:
+            num = int(r["detected"])
+        except (ValueError, TypeError):
+            continue
+        if num < 1:
+            out_of_range.append(r)
+        elif exp_start is not None and (
+            num < exp_start - 5 or num > exp_end + 5
+        ):
+            out_of_range.append(r)
+        else:
+            seen_nums.setdefault(num, []).append(r["pdf_page"])
+    return out_of_range, seen_nums
+
+
+def _build_analysis(
+    ocr_results: list, exp_start: int | None, exp_end: int | None
+) -> dict:
+    """Analyze ocr_results into the structure ``_build_issues`` expects.
+
+    Computes out-of-range, duplicate, sequence, and missing-page data from
+    the current page numbers without re-running OCR or opening the PDF.
+
+    :param ocr_results: Per-page OCR results (already finalized).
+    :param exp_start: Expected first page number, or None.
+    :param exp_end: Expected last page number, or None.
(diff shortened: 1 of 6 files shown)


grossir (approved):
Looks good! Just pending merge conflicts to solve

PR #108, opened 2026-06-25
fix(scanning): remove dismissed issue card and update issue count

### LINKED ISSUE #107: Dismissing a sidebar issue only dims the card, doesn't remove it or update the issue count
### Summary
In the step-1 process viewer sidebar, clicking **Dismiss** on an issue card dims the card but doesn't remove it from the list, and the "Issues (N)" count in the heading doesn't change. The dismissal does persist server-side (a reload shows the card gone and the count corrected), so this is a client-side UI bug only.

### Steps to reproduce
1. Open a scan with dismissable issues, e.g. `/scans/1872/process/` (step 1).
2. Click **Dismiss** on a warning/info issue card.
3. The card dims to ~30% opacity but stays in the list; "Issues (N)" is unchanged.
4. Reload, the card is gone and the count is correct.

### Expected
A dismissed issue is removed from the list and the "Issues (N)" count decrements immediately (and when the last one goes, the "All clear, no issues" state shows).

### Actual
The card is only set to `opacity: 0.3; pointer-events: none`; it is never removed and the count never updates without a reload.

### Root cause
The active handler is `window.dismissIssue` in `scanning/static/scanning/viewer_step1.js:768`; on success it only dims the card. The count is server-rendered (`scanning/templates/scanning/scan_process.html:80`, `Issues ({{ issues|length }})`) and no JS touches it.

quevon24:
Closes #107

Dismissing a step-1 sidebar issue only dimmed the card to 30% opacity. It stayed in the list and the "Issues (N)" heading never changed until a full reload. The dismissal did persist server-side, so this was a client-side UI bug.

## Changes
- Remove the card from the DOM on a successful dismiss and update the "Issues (N)" heading immediately.
- Swap in the existing "All clear, no issues" block once the last issue is dismissed.
- Surface server errors: the dismiss endpoint returns `{status: "error"}` when there are pending changes, which the old handler ignored silently. It now alerts the user and keeps the card.
- Consolidate the two duplicate `dismissIssue` handlers (`viewer_step1.js` and `viewer_sidebar.js`) into the single step-1 copy, and drop the dead `.issues-panel` / `.all-clear` lookups that referenced elements not present in the template.

## Template hooks
Added `id="issues-section"`, an `id="issues-count"` span in the heading, and an always-present (hidden) `id="issues-all-clear"` block so the JS can toggle state without a reload.

### DIFF
--- a/scanning/static/scanning/viewer_sidebar.js
+++ b/scanning/static/scanning/viewer_sidebar.js
@@ -436,29 +436,6 @@ function deleteDuplicates(btn) {
     });
 }
 
-function dismissIssue(btn, issueId) {
-    var cfg = window.SCAN_CONFIG;
-    fetch("/scans/" + cfg.docId + "/dismiss-issue/", {
-        method: "POST",
-        headers: {
-            "Content-Type": "application/json",
-            "X-CSRFToken": cfg.csrfToken,
-        },
-        body: JSON.stringify({ issue_id: issueId }),
-    })
-        .then(function (r) {
-            return r.json();
-        })
-        .then(function (data) {
-            if (data.status === "ok") {
-                btn.closest(".issue-card").style.opacity = "0.3";
-                btn.closest(".issue-card").style.pointerEvents = "none";
-            } else {
-                alert(data.message || "Cannot dismiss");
-            }
-        });
-}
-
 function pairOpinions() {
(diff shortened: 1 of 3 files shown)


grossir (approved):
Looks good!

PR #109, opened 2026-06-25
perf(scanning): stream upload to S3 in one pass, drop local disk write

### LINKED ISSUE #94: Move large-PDF disk write + S3 upload out of the queue_upload request
`scanning.views.queue_upload` does heavy, multi-minute I/O synchronously inside the scan upload HTTP request. The uploader / user is waiting while this happens:

1. Copy the uploaded file to output_dir/original_name (full pass over 1–2 GB to local disk).
2. `s3_sync.upload_file_to_s3()` -> boto3.client("s3").upload_file(...):  a second full pass, pushing the file to S3.

We should research options to improve this

quevon24:
Closes #94

`queue_upload` did two full passes over the uploaded PDF in prod after the browser finished sending: a local disk write, then a read-back to upload to S3. The web and daemon containers don't share disk, so the local copy is never read in prod (the daemon pulls the original from S3), making that pass pure waste.

This streams the file Django already spooled straight to S3 in a single pass and drops the local write in prod. The dev path (local FileField storage, shared with the daemon) is unchanged.

## Changes
- New `s3_sync.upload_fileobj_to_s3`: uploads the spooled `TemporaryUploadedFile` by path (or the file object for small in-memory uploads), tags it `application/pdf`, and logs the transferred size and duration.
- `queue_upload`: in prod, stream to S3 via the new helper; the local disk write now happens only on the dev path.
- Fixed stale comments and the `pdf_path` docstring that described the prod local copy as present after upload.

## Scope and follow-up
This removes the wasted disk pass and the 1-2 GB temp files the web container was writing, but the S3 upload itself still happens inside the request, so a large file still occupies the gunicorn worker (and the `--timeout 180` ceiling) during transfer. The size/duration log added here will surface the real prod cost.

### DIFF
--- a/scanning/models.py
+++ b/scanning/models.py
@@ -650,8 +650,8 @@ class Scan(AbstractDateTimeModel):
 
         Resolution order:
 
-        1. ``output_dir/<name>.original.pdf`` (present after upload or
-           after pulling from S3).
+        1. ``output_dir/<name>.original.pdf`` (present in DEV after
+           upload, or in prod after pulling from S3).
         2. Django ``FileField.path`` if the file actually exists on disk
            (covers DEV and tests where the FileField was written to
            MEDIA_ROOT).
--- a/scanning/s3_sync.py
+++ b/scanning/s3_sync.py
@@ -17,6 +17,7 @@ behavior as prod.
 from __future__ import annotations
 
 import logging
+import time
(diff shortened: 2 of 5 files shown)


grossir (approved):
LGTM

PR #112, opened 2026-06-26
fix(scanning): correct over-counted opinion coverage span

### LINKED ISSUE #111: fix(scanning): compute_coverage_gaps over-counts covered pages and masks real gaps
### Summary
`compute_coverage_gaps` (`scanning/utils.py:113-123`) inflates each opinion's covered page span, so genuinely uncovered pages are silently treated as covered and no gap warning is emitted.

### Location
`scanning/utils.py:114-123`

```python
for op in opinions:
    cp = op.get("caption_page", 0) + (start_page or 1)
    kp = (
        op.get("key_page", 0)
        + (start_page or 1)
        + op.get("page_count", 1)   # <-- double-counts the opinion length
        - 1
    )
    for p in range(cp, kp + 1):
        covered.add(p)
```

### Why it's wrong
An opinion spans `caption_page … key_page` inclusive, and blackletter sets `page_count == key_page - caption_page + 1`. The end of the covered span should just be `key_page + start_page`. Instead the code adds *both* `key_page` *and* `page_count`, so the span is extended by roughly `page_count - 1` extra pages past the real end. Because every opinion's span is inflated, the inflation bleeds into following opinions and masks almost any gap between them.

### Concrete example (opinion "Chevron U.S.A. v. University of....")
- `caption_page=1360`, `key_page=1371`, `page_count=12`, `start_page=1`
- Correct covered span: pages `1361 … 1372`
- Buggy covered span: `kp = 1371 + 1 + 12 - 1 = 1383` → pages `1361 … 1383`

When that opinion was mispaired (a false KEY_ICON ended it early), pdf_index 1372 was genuinely orphaned and should have surfaced as `Pages 1373-1373 (1 page) not covered`. The over-count swallowed it, so the unmatched-key warning was the only signal. (Surfaced while investigating the scan 1876 false-positive key icon, which itself was not a code bug.)

### Impact
- Coverage-gap warnings rarely fire, even when pages are dropped from all opinions.
- Affects both callers: the step-2 sidebar warnings (`views_process.py:389`) and the re-pair API endpoint (`views_api.py:321`).

quevon24:
## Summary
`compute_coverage_gaps` inflated each opinion's covered span by adding both `key_page` and `page_count`. Since `page_count == key_page - caption_page + 1`, the span end was pushed roughly `page_count - 1` pages past the opinion's real end, bleeding into the following opinions and masking genuinely uncovered pages, so no gap warning was emitted.

## Fix
Use the opinion's actual last page as the covered-span end: `page_end` when present, otherwise `key_page` (consistent with how `views_process.py` derives the opinion end via `ep = op.get("page_end", op.get("key_page", cp))`). The `page_count` term is dropped entirely.

## Tests
Added `scanning/tests/test_utils.py` for `compute_coverage_gaps`:
- regression for #111: an orphaned page between two opinions is reported as a gap, and an inflated `page_count` no longer swallows it
- `page_count` does not extend coverage; `page_end` does
- gaps before the first opinion, after the last, and multiple disjoint gaps
- `start_page` offset handling and empty/missing inputs

## Verification
When an opinion ends before the next one begins, the orphaned page now surfaces in the step-2 sidebar as a `Pages X-Y (N pages) not covered by any opinion` warning, which was previously masked:

sidebar

Closes #111

### DIFF
--- a/scanning/tests/test_utils.py
+++ b/scanning/tests/test_utils.py
@@ -1,0 +1,73 @@
+"""Tests for scanning.utils helpers."""
+
+from django.test import TestCase
+
+from scanning.utils import compute_coverage_gaps
+
+
+def _op(caption_page, key_page, **extra):
+    """Build a minimal opinion dict for coverage-gap tests.
+
+    :param caption_page: 0-based pdf index where the opinion starts.
+    :type caption_page: int
+    :param key_page: 0-based pdf index of the opinion's key icon (end).
+    :type key_page: int
+    :param extra: Additional opinion keys (e.g. page_end, page_count).
+    :returns: An opinion dict.
+    :rtype: dict
+    """
+    return {"caption_page": caption_page, "key_page": key_page, **extra}
+
+
+class TestComputeCoverageGaps(TestCase):
+    def test_no_gaps_when_fully_covered(self):
+        opinions = [_op(0, 3)]
+        self.assertEqual(compute_coverage_gaps(opinions, 1, 4), [])
+
+    def test_orphan_page_between_opinions_is_a_gap(self):
+        # Regression for #111: an opinion's covered span must end at its
+        # key_page, not key_page + page_count. A large page_count must not
+        # inflate coverage and swallow the orphaned page between opinions.
+        opinions = [
+            _op(0, 10, page_count=11),  # covers pages 1-11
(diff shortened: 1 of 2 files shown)


grossir (approved):
LGTM

Issue #113, opened 2026-07-03
Investigate scanning-web WORKER TIMEOUT freezes — add instrumentation to identify the blocking code path


grossir:
## Problem

`scanning-web` pods intermittently become unresponsive ("freeze") and drop in-flight requests. We have not yet identified the responsible code path. This issue records what is established, what is not, and the instrumentation needed to pin it down before attempting a fix.

## What we know (with evidence)

1. **Web workers time out repeatedly.** Sentry `SCANNING-1V`: `WORKER TIMEOUT (pid:…)`, 57 occurrences, ongoing, `logger: gunicorn.error`, `level: fatal`, `server_name: scanning-web-*`, spread across all hours over the last 30 days.
2. **Web workers are also killed by SIGKILL.** Sentry `SCANNING-1W`: `Worker (pid:…) was sent SIGKILL! Perhaps out of memory?`, 19 occurrences, ongoing, on `scanning-web-*`. ("Perhaps out of memory" is gunicorn's own guess on an externally-sent SIGKILL — see "What we don't know".)
3. **Server config** (`docker/django/docker-entrypoint.sh:29-40`): gunicorn, `--worker-class scanning.workers.UvicornWorker` (**async**), `--workers 4`, `--timeout 180`, `--max-requests 500 --max-requests-jitter 100`.
4. **Structural consequence of (3):** an async uvicorn worker signals liveness to the gunicorn arbiter from its event loop. A normal `await` keeps the loop ticking. Therefore a 180-second `WORKER TIMEOUT` means **the event loop was blocked by a synchronous call** (CPU-bound holding the GIL, or a synchronous blocking I/O call). This is a deduction from the worker model, not a measurement — but it constrains the cause to *synchronous work on the request path*.
5. **No per-request attribution exists today.** `WORKER TIMEOUT` events carry no Python stack (they're logged by the gunicorn master). Sentry performance tracing is **not enabled** — a spans query for `span.op:http.server` returns nothing — so we have no transaction/endpoint/duration data.
6. **Resource/probe facts** (from live manifests):
   - Web pods: `cpu request 500m`, **no CPU limit**; `memory request 500Mi`, **no memory limit**; QoS `Burstable`; spread one-per-node via `topologySpreadConstraints`.
   - Liveness probe is a bare TCP `socket.connect` — it **cannot detect a blocked event loop** (the port stays bound), so a wedged worker is never restarted on that basis.
[...]

PR #114, opened 2026-07-03
feat(scanning): report worker-timeout stack traces to Sentry

grossir:
## What & why

`scanning-web` workers time out (SCANNING-1V `WORKER TIMEOUT`, 57×) with **no attribution**: the timeout is logged by the gunicorn master and carries no Python stack, so the blocking code path is unknown. This PR makes a timed-out worker report *what it was doing* to Sentry, per the investigation plan in #113.

Since all views are sync `def` running in the asgiref threadpool, a CPU-bound call there can hold the GIL and starve the event-loop callback that pings gunicorn — which is what trips the 180s `WORKER TIMEOUT`.

## Change (`scanning/workers.py`)

**Re-wire SIGABRT → Sentry.** gunicorn's arbiter sends SIGABRT to a worker that misses its `--timeout` heartbeat. gunicorn's base worker would route that to the `worker_abort` hook, but uvicorn's worker resets all signals to `SIG_DFL` and only re-installs SIGUSR1 — so under this worker class `worker_abort` never fires and the process dies with no Python running. We restore a SIGABRT handler that captures **all thread stacks** into a `level="fatal"` Sentry event (`all_thread_stacks` extra), then chains to gunicorn's default abort. Best-effort (runs on the main thread, needs the GIL); a reporting failure is printed to stderr rather than swallowed.

**Enable faulthandler** for native-crash coverage: if a C extension such as fitz/MuPDF faults on a bad PDF (SIGSEGV/SIGBUS/etc.), all thread stacks are dumped to stderr/pod logs before the process dies, instead of a silent exit. Zero cost until an actual crash.

## Changed from the first revision (addressing review)
- **Dropped** the `faulthandler.dump_traceback_later(..., repeat=True)` watchdog — it's an unconditional periodic timer, not a wedge detector, so it would have flooded healthy pods with all-thread dumps every 120s. Kept only `faulthandler.enable()` for crash dumps.
- **Dropped** the gunicorn `--access-logformat`/`%(D)s` change — uvicorn's worker ignores gunicorn's access-log format (it emits a hardcoded format via its own protocol layer), so the duration atom would never have appeared. Per-request duration is better done in middleware and is deferred to #115.

## Notes / limitations
- The SIGABRT→Sentry report needs the GIL, so a worker fully wedged in GIL-holding C code may die before it runs; in practice fitz/MuPDF releases the GIL during heavy work, so it usually gets a slice to report.
- Native crashes reach pod logs only, not Sentry (Python can't run after a fault).
- **Not** included: Sentry performance tracing, and observability (memory/OOM watchdog, in-flight-request logging, concurrency counters) — tracked in #115.

Refs #113

### DIFF
--- a/scanning/workers.py
+++ b/scanning/workers.py
@@ -37,13 +37,17 @@ class UvicornWorker(BaseUvicornWorker):
         every signal to ``SIG_DFL`` here and only re-installs SIGUSR1, so under
         this worker class SIGABRT would otherwise kill the process with no
         Python running and nothing sent to Sentry. Restore a handler that
-        captures the current stacks first. Best-effort: it runs on the main
-        thread and needs the GIL, so a worker fully wedged in GIL-holding C
-        code may die before it runs — but in practice fitz/MuPDF releases the
-        GIL during heavy work, so the handler usually gets a slice to report.
+        captures the current stacks first. The Python handler is best-effort:
+        it runs on the main thread and needs the GIL, so a worker fully wedged
+        in GIL-holding C code may die before it runs. To cover that case we
+        also re-arm faulthandler for SIGABRT with ``chain=True``: its C-level
+        handler dumps all thread stacks to stderr (pod logs) *without* the GIL,
+        then chains to the Python ``handle_abort`` for the Sentry report once
+        the GIL is available.
         """
         super().init_signals()
         signal.signal(signal.SIGABRT, self.handle_abort)
+        faulthandler.register(signal.SIGABRT, chain=True)
 
     def handle_abort(self, sig: int, frame: Any) -> None:
         self._report_timeout_to_sentry()
@@ -67,7 +71,11 @@ class UvicornWorker(BaseUvicornWorker):
                     "gunicorn WORKER TIMEOUT: worker aborted after --timeout",
                     level="fatal",
                 )
-            sentry_sdk.flush(timeout=2.0)
+            # gunicorn's arbiter escalates SIGABRT to SIGKILL on its next
+            # murder-loop cycle (~1s later). Keep the flush sub-second so the
+            # event has a chance to ship before SIGKILL and so it doesn't stall
+            # the graceful exit (worker_abort + sys.exit(1)) in handle_abort.
+            sentry_sdk.flush(timeout=0.7)
         except Exception:
             # Never let reporting crash the abort/exit path, but don't swallow
             # silently: print to stderr (pod logs), which is signal-safe unlike
--- a/scanning/workers.py
+++ b/scanning/workers.py
@@ -1,3 +1,7 @@
+import faulthandler
+import signal
+import sys
+import traceback
 from typing import Any
 
 from uvicorn.workers import UvicornWorker as BaseUvicornWorker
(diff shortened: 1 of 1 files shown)


grossir:
@claude review

scanning/workers.py
```diff
@@ -9,3 +13,67 @@ class UvicornWorker(BaseUvicornWorker):
...
+        every signal to ``SIG_DFL`` here and only re-installs SIGUSR1, so under
+        this worker class SIGABRT would otherwise kill the process with no
+        Python running and nothing sent to Sentry. Restore a handler that
+        captures the current stacks first. Best-effort: it runs on the main
+        thread and needs the GIL, so a worker fully wedged in GIL-holding C
+        code may die before it runs — but in practice fitz/MuPDF releases the
+        GIL during heavy work, so the handler usually gets a slice to report.
+        """
+        super().init_signals()
+        signal.signal(signal.SIGABRT, self.handle_abort)
```
quevon24, line 53: This override replaces `faulthandler`'s SIGABRT handler with a Python one, which is weaker in exactly the scenario this PR targets. `faulthandler`'s handlers are C-level and dump all-thread stacks **without acquiring the GIL** — so a SIGABRT that arrives while fitz/MuPDF holds the GIL (the classic `WORKER TIMEOUT` hang) would still land in stderr/pod logs under faulthandler. `self.handle_abort` can't run until the GIL is free, so for a fully GIL-wedged worker we now get **nothing** — no Sentry event *and* no stderr dump (the `faulthandler.enable()` SIGABRT registration in `init_process` is already reset by `super().init_signals()` above).

We can keep both: register faulthandler for SIGABRT with `chain=True` right after installing the Python handler. faulthandler dumps GIL-free first, then chains to `handle_abort` for the Sentry report when the GIL is available.

```suggestion
        super().init_signals()
        signal.signal(signal.SIGABRT, self.handle_abort)
        # Keep a GIL-free fallback: faulthandler's C-level handler dumps all
        # thread stacks to stderr even when native code holds the GIL (the
        # exact WORKER TIMEOUT case the Python handler can't cover). chain=True
        # runs handle_abort afterwards for the Sentry report when the GIL frees.
        faulthandler.register(signal.SIGABRT, chain=True)
```

scanning/workers.py
```diff
@@ -9,3 +13,67 @@ class UvicornWorker(BaseUvicornWorker):
...
+                dump.append("".join(traceback.format_stack(thread_frame)))
+            stacks = "\n".join(dump)
+
+            with sentry_sdk.new_scope() as scope:
+                scope.set_extra("all_thread_stacks", stacks)
+                sentry_sdk.capture_message(
+                    "gunicorn WORKER TIMEOUT: worker aborted after --timeout",
+                    level="fatal",
+                )
+            sentry_sdk.flush(timeout=2.0)
```
quevon24: A 2 s flush likely exceeds gunicorn's abort→kill window. The arbiter's murder loop iterates roughly once per second; after sending SIGABRT it escalates to `SIGKILL` on the **next** cycle (~1 s later). If a flush actually needs close to 2 s — the slow-network case where the cap matters — SIGKILL arrives mid-flush, the event is dropped anyway, and meanwhile the blocking flush delays `super().handle_abort()` from running `worker_abort` + `sys.exit(1)` for a clean exit.

A sub-second timeout has a better chance of delivering the event before SIGKILL, and keeps the handler from stalling the graceful exit path.

```suggestion
            sentry_sdk.flush(timeout=0.7)
```

quevon24 (approved):
Thanks for addressing my comments. LGTM

PR #117, opened 2026-07-06
fix(scanning): retry transient daemon DB connection failures (SCANNING-20)

### LINKED ISSUE #116: Daemon: transient Postgres SSL connection errors (SCANNING-20)
## Summary

Sentry [SCANNING-20](https://freelawproject.sentry.io/issues/7515179433/) fires transient DB connection errors from the daemon:

```
OperationalError: connection failed: connection to server at "172.30.1.92", port 5432 failed:
SSL error: unexpected eof while reading
```

- **Culprit:** `scanning.management.commands.process_next_scan._process_next` (`process_next_scan.py:83`)
- **Occurrences:** 17 in ~5 weeks, **0 users impacted**
- **Handled:** yes — `run_daemon.py:118` catches it, logs via `logger.exception`, and the next tick reconnects fine.

## Root cause

Every daemon tick (~5s), `_process_next()` calls `connections.close_all()` (`process_next_scan.py:81`) and then `transaction.atomic()` (line 83), which forces Django to open a **brand-new** TCP+TLS connection (`CONN_MAX_AGE: 0` + `close_all()` means connections are never reused — we open a fresh one ~17k times/day).

Occasionally that fresh connect loses the TLS handshake with `SSL error: unexpected eof while reading` — i.e. the peer closed the socket without a clean TLS `close_notify`. Classic environmental causes: Postgres/pgbouncer restart or failover, a connection-limit rejection, or a network middlebox resetting the socket.

This is **environmental, self-recovering noise**, not a logic bug. The Sentry event is generated by the logging integration off the `logger.exception` call in the daemon loop (`mechanism: logging`, `handled: yes`).

## Proposed fix

1. **Retry connection acquisition in `_process_next`** — a transient connect failure should be retried a couple times with a short backoff before giving up on the tick. Eliminates virtually all of these events.

2. **Downgrade transient `OperationalError` to `logger.warning` in the daemon loop** (`run_daemon.py:120-121`) so a recovered blip doesn't create a Sentry error event, while genuinely unexpected exceptions still report as errors.

Sketch:

```python
# process_next_scan.py — _process_next()
connections.close_all()

scan = None
for attempt in range(3):
    try:
        with transaction.atomic():
            scan = (
                Scan.objects.select_for_update(skip_locked=True)
                .filter(status=Status.QUEUED)
                .order_by("date_created")
                .first()
            )
            if scan is None:
                return
            action = scan.queued_action or QueuedAction.FULL_PIPELINE
            Scan.objects.filter(pk=scan.pk).update(
                status=Status.PROCESSING,
                processed_at=timezone.now(),
                progress_message=f"Starting {action}...",
                progress_current=0,
                progress_total=0,
            )
        break
    except OperationalError as exc:
        connections.close_all()
        if attempt == 2:
            logger.warning(
                "DB connection failed claiming next scan after "
                "%d attempts: %s", attempt + 1, exc,
            )
            return
        time.sleep(0.5 * (attempt + 1))
```

Plus a test that raises `OperationalError` on the first attempt and asserts the retry succeeds.

_Fixes SCANNING-20_

quevon24:
## Summary

Sentry [SCANNING-20](https://freelawproject.sentry.io/issues/7515179433/) fires transient DB connection errors from the daemon:

    OperationalError: connection failed: connection to server at "172.30.1.92", port 5432 failed: SSL error: unexpected eof while reading

17 occurrences in ~5 weeks, 0 users impacted, already handled and self-recovering — but each one lands as a Sentry *error*.

## Root cause

`CONN_MAX_AGE: 0` plus `connections.close_all()` on every ~5s daemon tick means the daemon opens a brand-new TCP+TLS connection every tick (~17k/day) and never reuses one. Occasionally that fresh handshake loses `close_notify` (a Postgres/pgbouncer restart or failover, a connection-limit rejection, or a middlebox reset), raising `OperationalError`. It's environmental, self-recovering noise — the next tick reconnects fine.

It surfaces as a Sentry *error* only because the daemon loop logs it via `logger.exception`, and Sentry's `DjangoIntegration` promotes ERROR-level logs to events (default `event_level=ERROR`).

## Changes

- **Retry the DB-claim portion of a tick.** `process_next_scan._process_next()` was split into `_claim_next()` (the atomic claim) and `_dispatch()` (the pipeline run). The claim — `_recover_stale()` + `_claim_next()`, both idempotent under re-run — is now wrapped in `_claim_next_scan_with_retry()`, which opens a fresh connection and retries up to 3 times with a short backoff (0.5·attempt s). The long-running pipeline dispatch stays **outside** the retry so it runs exactly once with its existing error handling.
- **Downgrade recovered blips to WARNING.** A transient failure that survives all retries is logged at WARNING (a Sentry breadcrumb, not an error event), and the tick gives up cleanly until the next one.
- **Defense-in-depth in the daemon loop.** `run_daemon` now catches `OperationalError` from any scheduled task and logs it at WARNING before the generic `except Exception: logger.exception`, so a transient DB blip from `cleanup_processing_tmp` (or any path outside the retry) also stays out of Sentry's error stream.

## Tests

- New `scanning/tests/test_process_next_scan.py` (the command previously had no dedicated tests): retry-then-succeed, exhausted-retries (warns, scan stays QUEUED, no exception propagates), and empty-queue no-op.
- `scanning/tests/test_run_daemon.py`: added a case asserting a transient `OperationalError` is logged at WARNING, never ERROR.

All 261 tests pass (1 skipped).

Fixes: #116

### DIFF
--- a/scanning/management/commands/process_next_scan.py
+++ b/scanning/management/commands/process_next_scan.py
@@ -16,10 +16,18 @@ Examples:
 """
 
 import logging
+import time
 
 from django.core.management.base import BaseCommand
 
 logger = logging.getLogger(__name__)
+
+# Every tick opens a fresh TCP+TLS connection (CONN_MAX_AGE=0 +
+# connections.close_all()), so an occasional transient connect failure
+# (SSL close_notify loss on a Postgres/pgbouncer blip) is expected. Retry
+# a couple of times before giving up on the tick. See issue #116.
+MAX_DB_RETRIES = 3
+RETRY_BACKOFF_SECONDS = 0.5
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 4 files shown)


grossir (approved):
LGTM!

Issue #115, opened 2026-07-03
Add web-pod observability: memory/OOM watchdog, in-flight-request logging, concurrency counters


grossir:
## Problem

PR #114 (part of #113) captures the stack of a timed-out worker and logs per-request durations. That closes the **hang** attribution gap, but three diagnostic blind spots remain — all observability concerns that were deliberately scoped out of #114 to keep it low-risk and reviewable.

### 1. OOM / SIGKILL is a black box (highest value)
`scanning-web` workers are SIGKILL'd (SCANNING-1W, "perhaps out of memory"), and SIGKILL is **uncatchable** — the only way to get evidence is to log *before* the kill. There is currently **no in-process memory signal at all**. Proposed:
- A lightweight background thread logging worker RSS every N seconds (a memory-over-time trace in pod logs, line up against the kill timestamp).
- A **pre-OOM watchdog**: when RSS crosses a soft threshold (e.g. 85% of the pod limit), dump all-thread stacks + RSS to stderr/Sentry. The only in-process route to "what was the worker doing when it ran out of memory."

### 2. No "which request + params" attribution
faulthandler names the *view function*, but not the request context (which `scan.pk`, user, URL params). And gunicorn's access log writes a line **on response completion**, so a request that hangs until the worker is killed **never logs** — the access log shows everything except the culprit. Proposed:
- An **in-flight-request logger**: middleware records `(thread, url, scan_pk, user, start_time)` on entry into a shared dict; a background thread logs any entry older than N seconds (with elapsed + context) *while it's still hung*, and clears it on completion. Names the frozen endpoint with its params, before the kill.

### 3. Threadpool / concurrency saturation is invisible
A leading hypothesis is asgiref-threadpool exhaustion / GIL starvation, but nothing logs it. Proposed:
- Periodically log in-flight request count and `threading.active_count()` (pairs naturally with #2's shared state).

## Optional
- `send_default_pii=True` in Sentry init to attach user/request details to error events (privacy tradeoff).
- Structured/JSON logs for searchability in the EKS log aggregator.

## Definition of done
- [ ] Worker RSS is visible over time in logs; a pre-OOM threshold dump exists, letting SCANNING-1W be classified (pod OOM vs node pressure).
- [ ] A hung request is logged with its endpoint + params (`scan_pk`, user) while still in flight.
- [ ] In-flight/active-thread counts are logged, enough to confirm or rule out threadpool saturation.

## Notes
[...]

PR #122, opened 2026-07-07
feat(scanning): attach in-flight requests to WORKER TIMEOUT Sentry event

quevon24:
## What & why

Follow-up to #121. That PR added an in-flight-request registry + monitor thread so we could attribute *which* request wedged a web worker. But the attribution only ever reaches the **pod logs** — it never makes it into the Sentry WORKER TIMEOUT event, which is where triage actually starts.

The reason is thread scoping. The monitor thread's `web-monitor hung request …` warnings and the per-view INFO breadcrumbs are emitted from **background / request-serving threads**, and Sentry breadcrumbs are **per-thread** (contextvar-scoped isolation scope; `threading.Thread` children don't inherit the parent context). The fatal event is captured on the **main thread** inside the SIGABRT handler (`_report_timeout_to_sentry`), so none of those breadcrumbs are visible at capture time.

## Evidence

Verified against the first post-#121 deploy event — **SCANNING-2C**, `gunicorn WORKER TIMEOUT: worker aborted after --timeout`, release `18b2a12`:

- The worker booted at 20:00:00 and timed out at 20:14:46 (~14 min later).
- Over that window the monitor should have logged `web-monitor inflight=…` ~80+ times.
- The event's breadcrumb trail is **6 lines total**: gunicorn boot, the one-time `web-monitor started`, and uvicorn start. The `web-monitor started` line survives *only* because it's logged on the main thread in `init_process`. Zero periodic monitor lines, zero `hung request`, zero view breadcrumbs.

So #121's Sentry payload is effectively empty; the useful data lives only in stdout.

## Change

The registry (`observability._registry`) is a **lock-guarded module global**, not thread-scoped, so the main thread can read it directly. This adds:

- `observability.snapshot_for_report(now=None)` — returns a JSON-serializable list of in-flight requests (`thread`, `method`, `path`, `scan_pk`, `user`, `elapsed_s`), longest-running first. Reads only the lock-guarded registry, so it's safe to call from the SIGABRT handler; the main thread never holds `_registry_lock`, so it can't self-deadlock.
- `_report_timeout_to_sentry()` snapshots it and attaches it as an `in_flight_requests` scope extra alongside `all_thread_stacks`.

Net effect: the wedging endpoint + `scan_pk` + user now land on the fatal event itself.

## Testing

`python manage.py test scanning.tests.test_observability` — 8 tests (adds `SnapshotForReportTest`: empty-registry and elapsed/ordering). ruff clean.

## Notes / scope

- Purely additive; no change to the existing stack-dump path, monitor, middleware, or flush budget.
- In the specific SCANNING-2C event the captured stacks showed no request thread wedged (only idle sentry threads + the sleeping monitor + the main thread in the abort handler), i.e. the registry can legitimately be empty at capture — `snapshot_for_report` returns `[]` then, which is fine. This wires up attribution for the common case where a heavy request *is* in flight.

## Related

- Refs #115, follows #121 (in-flight registry + monitor), builds on #114 (SIGABRT stack capture).

### DIFF
--- a/scanning/middleware.py
+++ b/scanning/middleware.py
@@ -28,7 +28,12 @@ class InFlightRequestMiddleware:
 
     def __call__(self, request: HttpRequest) -> HttpResponse:
         ident = threading.get_ident()
-        observability.register_request(ident, request.path, request.method)
+        # get_full_path() (not request.path) so the query string is captured —
+        # the params (e.g. ?page=15&dpi=300 on a crop render, ?step=) are exactly
+        # what attributes a heavy request when it wedges a worker.
+        observability.register_request(
+            ident, request.get_full_path(), request.method
+        )
         try:
             return self.get_response(request)
         finally:
--- a/scanning/observability.py
+++ b/scanning/observability.py
@@ -104,7 +104,7 @@ def snapshot_for_report(now: float | None = None) -> list[dict]:
     Only reads the lock-guarded registry — never request/DB state — so it is
     safe to call from any thread. The main thread never holds ``_registry_lock``
     (only request/monitor threads do, briefly), so this cannot self-deadlock in
-    the signal handler. ``now`` defaults to ``time.monotonic()``; the newest
+    the signal handler. ``now`` defaults to ``time.monotonic()``; the oldest
     (longest-running) request is listed first.
     """
     if now is None:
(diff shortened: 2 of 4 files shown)


grossir (approved):
LGTM

I added a small change to capture the full request path, including query params, instead of only the path

PR #123, opened 2026-07-08
Slim the container image with a multi-stage build

### LINKED ISSUE #119: Slim the container image to cut pull time and reduce scale-up/reschedule churn
## Problem

The prod image `freelawproject/scanning:<sha>-prod` is **~1.57 GB**. In an hour of `kubectl get events -n scanning`, image pulls on fresh nodes took **33s to 3m22s**:

```
Successfully pulled image "...:07b9f2c-prod" in 2m54.523s. Image size: 1577608451 bytes.
Successfully pulled image "...:07b9f2c-prod" in 3m22.877s ...
Successfully pulled image "...:07b9f2c-prod" in 1m58.966s ...
```

Because the cluster is churning nodes heavily (HPA scaling web 4↔8 on CPU, plus node taint-evictions / `NodeNotReady`), pods are frequently rescheduled onto nodes without the image cached. Each such move incurs minutes of unavailability, and the **startup probe fails while the pod is still pulling/booting**:

```
Startup probe failed: Get "http://.../monitoring/heartbeat/": context deadline exceeded
```

So a large image directly **amplifies churn**: slow pulls → longer unavailability windows → failed startup probes → more restarts/reschedules.

## Impact

- Minutes of reduced capacity on every scale-up and every node move.
- Startup-probe failures during normal autoscaling, adding to pod churn.
- Slower deploys and slower recovery from node loss.

## Root causes (from `docker/django/Dockerfile`)

1. **Full base image** — `FROM python:3.12` (not `-slim`); carries compilers and dev tooling into runtime.
2. **Single-stage build** — build-only tooling stays in the final image: `nodejs`, `npm`, the installed `node_modules`, `git`, `curl`, `uv`. `node_modules`/npm are only needed for `npm run build` (Tailwind) at build time.
3. **ML stack shipped to the web tier** — `uv sync --all-extras` installs the `blackletter[analyze]` deps (YOLO/Ultralytics, PaddleX, torch, tesseract). The **web pods don't run YOLO/OCR/generation — the daemon does** — yet both share this one image, so the web tier pays the weight (and torch/paddle are typically the dominant layers).

## Proposed remediation (roughly by impact/effort)

1. **Multi-stage build with a slim runtime stage.** Build in the full image, then `COPY --from=build /opt/venv /opt/venv` + app + built static into `FROM python:3.12-slim`. Leave `nodejs`, `npm`, `node_modules`, `git`, `curl`, `uv` behind; install only runtime apt libs (`libpq5` instead of `libpq-dev`, `tesseract-ocr(-eng)`, `libgl1`, `libglib2.0-0`).
2. **Measure before cutting.** Run `docker history` / `dive` on the current image to confirm which layers dominate (torch vs paddle vs base) and target the biggest first.
3. **Trim the ML weight for web.** If torch/paddle dominate: pin **CPU-only** wheels, and/or split the `analyze` extras out of the web image (separate build target or image) so only the daemon carries them. Biggest single win if ML deps are the bulk.
4. **Operational mitigations (parallel, low-risk):** loosen the **startupProbe** (`failureThreshold`/`periodSeconds`) so a slow first boot behind an image pull doesn't fail startup; consider pre-pulling/warming the image on new nodes (e.g. a warmer DaemonSet) so reschedules hit cache.

## Acceptance criteria

- [ ] Runtime image measurably smaller (target: **< ~700 MB**, or a documented number after measuring).
- [ ] Cold-node pull time down proportionally (target: **< ~45s**).
- [ ] No `Startup probe failed` events attributable to pull/boot time during a normal HPA scale-up.
- [ ] Web and daemon behavior unchanged (daemon still has the full ML stack).

## Notes / relates to

- Surfaced while investigating web-pod churn for #113 (and the observability work in #115). The image size is an **aggravator** of churn, not the root cause — the churn itself (HPA flapping, node taint-evictions, worker-saturation liveness kills) is tracked separately.

quevon24:
Closes #119

## What

Rework `docker/django/Dockerfile` into a multi-stage build with a slim runtime, so the prod image stops shipping build-only tooling and dev dependencies. This cuts pull time on cold nodes and reduces the churn amplification described in the issue.

## Changes

- **Multi-stage build.** `build-base` (full `python:3.12` + compilers/node/uv) feeds three stages: `deps-prod`/`deps-dev` build the venv at `/opt/venv`, `static-builder` runs `npm run build` for Tailwind. The default `prod` target is `python:3.12-slim` and copies in only the venv, app source, and compiled CSS. `nodejs`, `npm`, `node_modules`, `git`, `curl`, `uv`, and compilers are left behind.
- **Drop the dev extra from prod.** Prod now runs plain `uv sync` (no `--all-extras`), so `mypy`, `pytest`, `pre-commit`, `django-debug-toolbar`, and `django-extensions` no longer ship to prod.
- **Trim runtime apt.** `libpq5` (not `libpq-dev`), `tesseract-ocr(-eng)`, `libgl1`, `libglib2.0-0`, `libgomp1`. Dropped `postgresql-client` (unused in the prod entrypoint).
- **Fix a cache-permission crash.** `gunicorn`/the daemon run as `www-data`, whose home `/var/www` isn't writable. Several ML libs cache under `~` — notably paddle's `~/.cache/paddle`, which has no env override — so this crashed on import/first use. Set `HOME=/tmp` (matching the existing `/tmp` cache convention) and fix the no-op `PADDLEX_HOME` -> `PADDLE_PDX_CACHE_HOME`.
- **dev stage for local dev + CI.** `docker-compose.yml` and the CI test build now target the `dev` stage, which keeps `uv` and the dev deps needed when the suite runs with `DEVELOPMENT=True` (`django_extensions` is added to `INSTALLED_APPS` whenever `DEVELOPMENT` is true, independent of `TESTING`).

## Results (measured on a local build)

- Image size: **~1.57 GB -> ~1.07 GB** (~32% smaller).
- Full ML stack imports cleanly under the slim runtime as `www-data`: torch 2.11.0+cpu, paddle 3.3.1, cv2 4.10.0, ultralytics, paddleocr, ocrmypdf, pymupdf, blackletter, django. Confirms `libgomp1` + the trimmed apt set are sufficient and the daemon is unaffected.
- `manage.py check` as `www-data` with prod settings: `0 issues`.
- paddlex/paddle cache dirs resolve to writable `/tmp/paddlex` and `/tmp/.cache/paddle` — the previous `PermissionError: /var/www` is gone.

## Acceptance criteria status

- [x] Runtime image measurably smaller — **~1.07 GB** (documented number; see below re: the <700 MB target).
- [ ] **< ~700 MB not reached by this PR.** torch (742 MB) + paddle (717 MB) dominate the venv and the web tier still carries them. Hitting <700 MB needs proposal #3 — moving the `analyze`/ML deps into a daemon-only image. They're currently hard `dependencies` (not an extra), so it's a real refactor; suggest a follow-up issue.
- [x] Web and daemon behavior unchanged (daemon still has the full ML stack).
- [ ] Cold-node pull time / startupProbe items live in the infra repo's k8s manifests, out of scope here.

## Notes

- CPU-only torch/paddle wheels (proposal #3, first half) were already pinned in `pyproject.toml` — no change needed.
- Suggest opening an independent issue to split the image into separate web and daemon builds (proposal #3), so the web tier drops the ML stack (torch/paddle/etc.) and lands under the ~700 MB target. It's a self-contained follow-up: carve the ML deps into an optional group, keep them daemon-only, and guard in CI that the web image boots without them.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,8 +26,9 @@ jobs:
           push: false
           tags: scanning:latest
           outputs: type=docker,dest=/tmp/scanning.tar
-          build-args: |
-            BUILD_ENV=dev
+          # Test the dev stage: the suite runs with DEVELOPMENT=True, which
+          # pulls in dev-only apps (django_extensions, ...) absent from prod.
+          target: dev
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
--- a/docker/django/Dockerfile
+++ b/docker/django/Dockerfile
@@ -136,11 +136,14 @@ WORKDIR /opt/scanning
 COPY --from=deps-prod /opt/venv /opt/venv
 
 # App source (build context excludes node_modules and built static via
-# .dockerignore) plus the compiled Tailwind CSS from the static stage.
+# .dockerignore) plus the full compiled static-global tree from the static
+# stage. That tree (Tailwind CSS + FontAwesome CSS + webfonts) is gitignored,
+# so it is absent from a fresh checkout's build context and must come from the
+# builder, or collectstatic silently ships an image missing those assets.
 COPY . .
 COPY --from=static-builder \
-     /opt/scanning/scanning/assets/static-global/css/tailwind_styles.css \
-     /opt/scanning/scanning/assets/static-global/css/tailwind_styles.css
+     /opt/scanning/scanning/assets/static-global/ \
+     /opt/scanning/scanning/assets/static-global/
 
 # Create directories for collectstatic output and uploaded/processed files
 RUN mkdir -p /opt/scanning/scanning/assets/static/ \
--- a/docker/django/Dockerfile
+++ b/docker/django/Dockerfile
@@ -1,3 +1,10 @@
+# syntax=docker/dockerfile:1
+
+##############################################################################
+# build-base: full image with compilers and build tooling.                   #
+# Shared parent for the dependency- and static-build stages. Nothing from     #
+# here ships to prod unless it is explicitly COPYed into the runtime stage.   #
+##############################################################################
 FROM python:3.12 AS build-base
 
 # Install apt dependencies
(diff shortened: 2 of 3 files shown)


grossir (approved):
LGTM. Let's keep watching the deployment to see if this helps

PR #125, opened 2026-07-08
fix(scanning): bound daemon re-queue loop with interruption cap

quevon24:
## What & why

The daemon re-queues an in-flight scan (`PROCESSING → QUEUED`) whenever it's killed (SIGTERM) or a scan times out mid-pipeline, and it deliberately does *not* consume the scan's RunPod retry budget — so a deploy or eviction can't burn a scan's retries.

The gap: that re-queue is unbounded. If the daemon pod is terminated repeatedly, the same scan can be re-queued forever — silently redoing GPU work on every attempt, never completing, and never surfacing anywhere. Sentry issue [SCANNING-1F](https://freelawproject.sentry.io/issues/7461261769/) shows the daemon receiving SIGTERM ~8×/day, so this loop is reachable in practice.

## The change

Track interruptions in a new `interruption_count` field, kept separate from `retry_count` (these are infra interruptions — deploys, evictions, OOM — not scan failures). Once a scan exceeds `DAEMON_MAX_INTERRUPTIONS` (default 5, env-overridable), it moves to a new terminal `ERROR_INTERRUPTED` status instead of being re-queued again.

Both re-queue paths feed the same guard: the SIGTERM handler in `run_daemon` and the stale-recovery path in `process_next_scan` (which covers SIGKILL/OOM, since those never run the signal handler).

Flagged scans are made visible rather than left silent:
- Admin gains an `interruption_count` column, an "interrupted too often" filter, and a re-queue action that resets the counter.
- The scan list shows a warning banner (mirroring the existing retry-cap banner) linking admins to the filtered changelist.

## A note on scope

This is a safety net to *catch* these cases and make them visible — not a fix for the underlying pod churn. Once the daemon deployment is running more stably, hitting the interruption cap should be rare. Separately, it's still worth a look at whether the daemon's scheduling/resources could be tuned to reduce how often it's terminated in the first place.

## Testing

`makemigrations` produced `0006_scan_interruption_count_alter_scan_status`. Full suite passes (275 tests), including new coverage for both re-queue paths: counter bump, flag-at-ceiling, re-queue-below-ceiling, and fresh-scan-left-alone.


scanning/models.py
```diff
@@ -580,6 +589,77 @@ class Meta:
...
+        )
+        if not pks:
+            return 0, 0
+
+        flag_message = (
+            f"Interrupted {max_interruptions}+ times without completing "
+            "(daemon killed or timed out mid-pipeline). Flagged for review."
+        )
+
+        # Single atomic UPDATE: increment interruption_count and branch on the
```
grossir, line 638: can you also logger.info() this so it appears on Sentry breadcrumbs?

scanning/management/commands/run_daemon.py
```diff
@@ -170,6 +174,12 @@ def _handle_signal(self, signum, frame):
                     "Re-queued %d in-flight scan(s) before shutdown.",
                     requeued,
                 )
+            if flagged:
+                logger.warning(
+                    "Flagged %d scan(s) as ERROR_INTERRUPTED after too many "
+                    "daemon interruptions.",
+                    flagged,
+                )
```
grossir: this seems important, maybe it's best to promote it to `logger.error`

grossir (commented):
This looks good

Just a few minor logger recommendations so we can get some more info on Sentry

Also, while reviewing this I thought about actually recovering the interrupted jobs. May be worth to implement? https://github.com/freelawproject/scanning/issues/127

quevon24:
> This looks good
> 
> Just a few minor logger recommendations so we can get some more info on Sentry
> 
> Also, while reviewing this I thought about actually recovering the interrupted jobs. May be worth to implement? #127

I would leave #127 for another PR, but I'll get to work on it.

grossir:
Yes @quevon24 let's leave that for another PR

PR #126, opened 2026-07-09
Large-upload 504 fix: preview-only viewer + direct-to-S3 upload (closes #124)

### LINKED ISSUE #124: Large uploads 504 at the gateway even though the file lands in S3 and the scan is created
## Problem

Uploading a large PDF (observed with a ~2 GB / 1383-page volume, e.g. `POST /queue/f-supp-3d/750/`) returns a **504 Gateway Timeout from CloudFront** to the browser:

```
504 Gateway Timeout ERROR
Generated by cloudfront (CloudFront) HTTP3 Server
```

But the upload actually **succeeds server-side**: the file lands in S3 and the scan is created and starts processing. Prod logs show the write completing normally:

```
INFO Streamed 2004.5 MB to s3://com-courtlistener-scanning-private-storage/processing/1938/f-supp-3d/750/1/f-supp-3d.750.1.1383.original.pdf for scan 1938 in 5.2s
```

So this is a **false failure**: the origin finishes the work, but the gateway already gave up on the response.

## Root cause

The upload is a single non-chunked `XMLHttpRequest` POST of the whole file (`queue_detail.html`), handled synchronously by a web worker:

1. Django's default `TemporaryFileUploadHandler` receives the entire body from the client and spools it to a temp file on the worker's disk. For a multi-GB file over a normal connection this is the slow part and it holds the worker the whole time.
2. `s3_sync.upload_fileobj_to_s3` then pushes that temp file to S3 in one pass — the `5.2s` in the log (server→S3, fast).
3. The scan is created and queued; gunicorn's `--timeout 180` is generous enough that the origin finishes.

The 504 comes from the gateway, not the app: CloudFront's origin-response timeout (30s default, 60s max without a quota increase) and the ALB idle timeout (60s default) are both far below the time it takes to receive a multi-GB body. gunicorn's higher 180s timeout is exactly why the origin keeps working and logs success *after* the browser has already received the 504.

## Upload wall-clock is bounded by client bandwidth

The timeout is not a fixed "this file is too big" line — it is `file_size / client_upload_bandwidth` vs. the gateway budget, so a slower connection (or a bigger file) is more likely to 504.

The logged `Streamed ... in Xs` time is misleading: it measures only the origin->S3 push (a timer around the `s3.upload_file` call), not the browser->server receive. In prod that push is server(AWS)->S3 in-region (~5s) and hides the slow leg; the browser->CloudFront transfer of the whole body (bounded by the uploader's uplink) happens earlier during body parsing and is never logged.

Real prod wall-clock for the request is roughly:

```
[browser -> server: ~2 GB over the uploader's uplink ~= 60-90s+]   <- dominant, unlogged
+ [server -> S3 push: ~5s]                                          <- the only part logged
+ [scan create + queue: ~0]
```

Concrete evidence: running the same 2 GB file locally logs `Streamed ... in 68.5s` — that 68.5s is the laptop->S3 push over a home uplink (~2004.5 MB / 68.5s ~= 29 MB/s ~= 234 Mbps), which is a direct measurement of the uploader's connection. In prod that same uplink is spent on the browser->CloudFront leg instead, pushing the total well past CloudFront's origin-response timeout (30s default, 60s max) and the ALB idle timeout (60s).

This is why direct-to-S3 (remediation #1) is the real fix: it does not make the bytes travel faster (still the same uplink), but S3 multipart uploads have no 60s ceiling, hold no Django worker, and never touch CloudFront's origin timeout, so a slow multi-GB upload completes instead of being cut off.

## Why it started showing up now

The deployed upload path is unchanged — `upload_fileobj_to_s3` (spool-then-push) predates all recent work, and `scanning/views.py` / `scanning/s3_sync.py` were not touched by any merge in the last several days. This is a **pre-existing latent limit**: this volume is simply the first upload large enough to cross the gateway timeout; smaller volumes finished the round trip in time.

Two things that are NOT the cause, ruled out during investigation:
- The multi-stage Docker slim change (#119 / PR #123) — build-only, does not touch request handling.
- The `streaming-s3-upload` prototype — unmerged/undeployed; the log line and S3 key format confirm prod runs the older spool-then-push path, not the streaming handler.

The web-pod observability work (#115) is likely **why this became visible** (its monitor now flags requests in flight past `WEB_INFLIGHT_WARN_SECONDS`), but it does not cause the timeout.

## Impact

- Broken UX: the user sees an error and never gets redirected to the process page, even though the upload worked.
- Duplicate-scan risk: because it looks like it failed, a volunteer re-uploads. With `new_scan=1` each retry creates another `Scan` and re-streams the whole file, spawning duplicates and extra daemon work.
- Gets worse as volume sizes grow.

## Proposed remediation (roughly by impact/effort)

1. **Direct-to-S3 presigned upload (durable fix).** The browser uploads straight to S3 (presigned POST / multipart), bypassing CloudFront/ALB/gunicorn entirely; the app just issues the presigned URL and records the key on a fast completion callback. Removes the gateway clock and the worker-occupancy problem in one move. (The `streaming-s3-upload` prototype was reaching for this but only removed the double-transfer — it still holds a worker for the whole receive, so it would not fix the 504.)
2. **Raise the gateway timeouts (quick stopgap, infra repo).** Bump the CloudFront origin-response timeout (→ 60s, quota increase for more) and the ALB idle timeout above the real end-to-end upload time. Low-risk and would stop the 504 for uploads that finish inside the new window, but a big/slow enough upload still times out. Not a real fix.
3. **Idempotent uploads (cheap interim guard).** Dedupe so a retry doesn't create a second scan — key the create on volume + page-range + user, or use the deterministic S3 key to detect "already uploaded." Blunts the duplicate-scan problem while the timeout issue exists.

## Acceptance criteria

- [ ] A multi-GB upload completes without the browser receiving a 504.
- [ ] A retried/duplicate upload does not create a second `Scan` for the same volume + page range.
- [ ] No web worker is held for the duration of a multi-GB transfer (direct-to-S3 path).

## Notes / relates to

- Surfaced while investigating the 504 on a 2 GB upload; made visible by the #115 web-pod observability work.

quevon24:
## What & why

Large scan uploads were 504ing at the gateway even though the file landed in S3 and the scan row was created (#124). The upload itself succeeded, but the very next page — `/scans/<pk>/process/` — hung for minutes and the PDF viewer failed with `net::ERR_CONTENT_LENGTH_MISMATCH`.

Two root causes, both around moving the multi-GB original PDF that has no business being on the request path:

1. **The process view eagerly pulled the scan's entire S3 processing prefix to local `/tmp` before returning any HTML.** Right after a direct-to-S3 upload the only object in that prefix is the original (up to 2 GB), so the pull took minutes — blocking a response that renders purely from DB fields and never reads those files.
2. **`serve_scan_pdf` fell back to streaming the original** when no processed PDF existed. Streaming a ~1.7 GB file ran past gunicorn's `--timeout 180`, so the worker was killed mid-stream; the browser received fewer bytes than the declared `Content-Length` → `ERR_CONTENT_LENGTH_MISMATCH`. It literally cannot succeed through the current stack, and a 2 GB file is unusable in pdf.js regardless.

This PR ships the presigned direct-to-S3 upload flow together with the fixes for the 504 it exposed.

## Changes

- **feat: presigned direct-to-S3 upload for scan originals** — the browser uploads the original straight to S3 via a presigned POST; `presign/` issues the policy (2 GB cap) and `confirm/` verifies the object and creates the scan.
- **perf: drop the eager S3 pull from the process view** — the page renders from the DB (`page_map`, `ocr_results`, `opinions_json`, detections, `redaction_rects`); the PDF and crop endpoints already pull what they need lazily.
- **perf: reuse a single boto3 S3 client** — `boto3.client("s3")` reloads the service model and reruns the credential chain on every call; cache one client and reuse it (thread-safe for calls).
- **fix: `serve_scan_pdf` serves only the small preview (OCR › bitonal), never the original** — a new targeted `download_preview_pdf` pulls only the bitonal/OCR PDF from S3 (skips the original and `images/`). When no preview exists yet, it returns HTTP 202 with a status-aware message, and the viewer shows "We're still processing this file…" and auto-retries until the preview lands, instead of a hard error. The original stays reachable only for server-side crops (`serve_original_crop`).

## Result

- A pending-review scan (bitonal exists) loads its small preview immediately.
- A still-processing scan shows a friendly message and swaps in the preview on its own once ready.
- The 2 GB original is never streamed to the browser — no more gateway timeout.

## Testing

- Full suite green: 272 passed, 1 pre-existing skip.
- Updated `serve_scan_pdf` tests to the new contract (202 not-ready, original never served) and added a `download_preview_pdf` selectivity test (pulls only bitonal/OCR, skips original + images).

Closes #124


scanning/settings/project/processing_storage.py
```diff
@@ -14,3 +14,9 @@
 PROCESSING_TMP_CLEANUP_INTERVAL_SECONDS = env.int(
     "PROCESSING_TMP_CLEANUP_INTERVAL_SECONDS", default=900
 )
+
+# How long a presigned direct-to-S3 upload (PendingUpload) may sit
+# unconfirmed before cleanup_processing_tmp deletes it -- and, if the
+# upload never landed, its fileless scan. Covers users who close the tab
+# mid-upload.
+PENDING_UPLOAD_TTL_HOURS = env.float("PENDING_UPLOAD_TTL_HOURS", default=24.0)
```
grossir: Should we use a shorter default here? `_sweep_pending_uploads` is ran on the `cleanup_processing_tmp` command, it seems every 15 minutes (900 secs) through `PROCESSING_TMP_CLEANUP_INTERVAL_SECONDS` (this is not overriden in prod). 

On the other hand, `S3_UPLOAD_PRESIGNED_TTL` is 1 hour. 

So, a reasonable time should be to wait at most 1 hour before `_sweep_pending_uploads` includes an upload in the checks?

grossir (commented):
Looks good! Only comment for now is about the cleanup sweep delay

Also, this will need to update the migration since we merged the interruption count migration

grossir (approved):
LGTM

PR #128, opened 2026-07-10
Split into lean web and heavy daemon images

quevon24:
## What

Splits the single scanning image into two, so the web tier stops shipping the ~2.5 GB ML stack it never runs (detection is offloaded to RunPod).

- **`pyproject.toml`**: `torch`, `torchvision`, `ultralytics`, `paddlepaddle`, `paddleocr`, `opencv-python`, and `blackletter[detect,analyze]` move out of base deps into a new **`local-ml`** extra. Base now depends on plain `blackletter` (+ pymupdf/pdfplumber/ocrmypdf/pytesseract), so the web tree carries no torch/paddle.
- **`Dockerfile`**: `deps-prod` → `deps-web` (lean, `uv sync`) + `deps-ml` (heavy, `uv sync --extra local-ml`); `prod` → **`prod-web`** (lean, drops `libgl1`/`libgomp1`, strips blackletter's bundled YOLO weights) + **`prod-ml`** (heavy). `dev` target keeps `--all-extras` so local dev can still run everything with `RUNPOD_ENABLED=False`.
- **`Makefile` / `deploy.yml`**: one Docker Hub repo, two variant tags — `:<sha>-web` and `:<sha>-daemon`. `deploy-web` uses `-web`, `deploy-daemon` uses `-daemon`, and the collectstatic/migrate temp pod uses `-web`.

## Image sizes (compressed, comparable to the current 1.46 GB)

| Image | Target | Size | vs current |
|---|---|---|---|
| web | `prod-web` | **~309 MB** | ~4.7× smaller |
| daemon | `prod-ml` | **~1.2 GB** | ~same (carries the full stack) |

The daemon stays heavy on purpose: it runs detection in-process whenever RunPod is off (no credits / offline / local dev), so it must keep torch/paddle.

## Verified

Built both targets locally. `prod-web` imports `blackletter.api`/`scanner`/`_pair_opinions` + `cv2` with **no** torch/paddle present; `prod-ml` imports torch 2.11.0+cpu, paddle 3.3.1, ultralytics. `docker compose up` is unchanged (both dev services still build `target: dev`).

## ⚠️ Do not merge until blackletter is released

This depends on **freelawproject/blackletter#58** (the base/detect/analyze split). Until it lands on PyPI, `[tool.uv.sources]` pins `blackletter` to that branch. Before merge:

1. Wait for blackletter 0.0.14 to be published.
2. Drop the `blackletter = { git = … }` line in `[tool.uv.sources]`.
3. Bump the `blackletter` specifiers to `>=0.0.14` (base) and in `local-ml`.
4. Re-lock (`uv lock`) and confirm both images still build.

Also note the `:<sha>-prod` tag is no longer produced — confirm no k8s manifest or external reference still points at `-prod`.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -87,12 +87,12 @@ jobs:
 
       - name: Launch Temporary Pod
         run: |
-          kubectl run temp-pod-${{ steps.vars.outputs.sha_short }} -n ${{ env.EKS_NAMESPACE }} --image=freelawproject/scanning:${{ steps.vars.outputs.sha_short }}-prod --restart Never --pod-running-timeout=120s --overrides='
+          kubectl run temp-pod-${{ steps.vars.outputs.sha_short }} -n ${{ env.EKS_NAMESPACE }} --image=freelawproject/scanning:${{ steps.vars.outputs.sha_short }}-web --restart Never --pod-running-timeout=120s --overrides='
           {
               "spec": {
               "containers": [{
                   "name": "temp-pod",
-                  "image": "freelawproject/scanning:${{ steps.vars.outputs.sha_short }}-prod",
+                  "image": "freelawproject/scanning:${{ steps.vars.outputs.sha_short }}-web",
                   "command": ["/bin/sh", "-c", "trap : TERM INT; sleep 259200 & wait"],
                   "envFrom": [{
                   "secretRef": {
@@ -140,7 +140,7 @@ jobs:
         run: aws eks update-kubeconfig --region ${{ env.AWS_REGION }} --name ${{ env.EKS_CLUSTER_NAME }}
 
       - name: Rollout scanning-web
-        run: kubectl set image -n ${{ env.EKS_NAMESPACE }} deployment/scanning-web web=freelawproject/scanning:${{ needs.setup-deploy.outputs.sha_short }}-prod
+        run: kubectl set image -n ${{ env.EKS_NAMESPACE }} deployment/scanning-web web=freelawproject/scanning:${{ needs.setup-deploy.outputs.sha_short }}-web
       - name: Watch scanning-web rollout status
         run: kubectl rollout status -n ${{ env.EKS_NAMESPACE }} deployment/scanning-web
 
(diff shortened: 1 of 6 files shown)


quevon24:
Follow-ups from the review:

- Fixed the `push-image` exit-status masking in `docker/django/Makefile` (chained the pushes with `&&`) in 768cdc4.
- The cv2 conflict is a blackletter bug, filed as freelawproject/blackletter#61. `blackletter[analyze]` installs both `opencv-python-headless` (base) and `opencv-python` (via ultralytics/doctr), which conflict on the `cv2` namespace. The fix (base -> `opencv-python`) has to land there and ship as 0.0.14. Once it does, this PR needs to: restore `libgl1`/`libgomp1` on `prod-web`, drop the now-redundant `opencv-python==4.10.0.84` pin from `local-ml`, drop the git pin, and re-lock. Blocking merge until then.

quevon24:
Update on the cv2 conflict.

The blackletter fix is up as freelawproject/blackletter#62 (base switches to `opencv-python`, closes freelawproject/blackletter#61). Once that ships as 0.0.14, base no longer pulls `opencv-python-headless`, so the dangerous 4.x/5.x mismatch is gone.

While resolving I found a second, pre-existing cv2 variant in the daemon tree: `paddleocr -> paddlex` requires `opencv-contrib-python`, which also owns the `cv2` namespace and coexists with the `opencv-python` that ultralytics/doctr/blackletter require. This can't be deduplicated (uv can't rename one distribution into the other, and both are hard requirements of third-party packages). To keep them safe together I added a `[tool.uv] override-dependencies` stanza pinning `opencv-python`, `opencv-contrib-python`, and `opencv-python-headless` all to `==4.10.0.84` (a2a9d68), and dropped the now-redundant `opencv-python` pin from the `local-ml` extra. Same OpenCV version means one shared ABI, so coexistence is safe even though it is still two wheels.

Updated pre-merge checklist (unchanged from before except the pin/override items):

1. Wait for blackletter 0.0.14 on PyPI.
2. Drop the `blackletter = { git = ... }` line in `[tool.uv.sources]`; bump the base and `local-ml` specifiers to `>=0.0.14`.
3. Restore `libgl1`/`libgomp1` on `prod-web` (base is non-headless again as of 0.0.14).
4. Re-lock (`uv lock`) so the override-dependencies take effect, and confirm the daemon lock has a single cv2 version and no `opencv-python-headless`.
5. Confirm both images still build.

grossir (approved):
LGTM

PR #133, opened 2026-07-15
Remove gunicorn worker recycling (max-requests)

### LINKED ISSUE #132: Disable gunicorn worker recycling (max-requests) to surface memory leaks
## Context

The infra team is debugging dying nodes in the scanning pods. Their recommendation: stop recycling gunicorn workers so memory issues become visible instead of being masked.

Our prod entrypoint currently recycles each worker every 500-600 requests ([docker-entrypoint.sh#L38](https://github.com/freelawproject/scanning/blob/main/docker/django/docker-entrypoint.sh#L38)):

```sh
--max-requests ${MAX_REQUESTS:-500} \
--max-requests-jitter 100 \
```

`max-requests` is explicitly a band-aid: gunicorn's docs describe it as "a simple method to help limit the damage of memory leaks." With recycling on, a slow leak resets before it ever shows in memory graphs. With long-lived workers, RSS climbs monotonically and can be correlated with node deaths. Our workers do heavy fitz/MuPDF work, and native C-extension allocations are the classic leak source that worker recycling quietly papers over.

## Options

1. **Set `MAX_REQUESTS=0` in the pod env (no code change).** The flag is already env-driven, and gunicorn 25.3.0 treats `max_requests=0` as disabled (the hardcoded jitter is ignored in that branch, and the value flows through our custom `UvicornWorker` as `limit_max_requests=sys.maxsize`). Reversible without an image rebuild, and infra can toggle it per environment during the investigation.
2. **Remove both flags from the entrypoint.** Permanent and explicit, but requires a rebuild to change and removes the ability to re-enable recycling via env if we decide we want it back after the investigation.

quevon24:
Fixes #132 (option 2).

Removes `--max-requests ${MAX_REQUESTS:-500}` and `--max-requests-jitter 100` from the `web-prod` gunicorn command so workers are long-lived. With recycling on, a slow leak resets before it ever shows in memory graphs; with long-lived workers, RSS climbs monotonically and can be correlated with the node deaths infra is investigating.

No code changes needed elsewhere: with the flag gone, gunicorn defaults `max_requests` to 0, and uvicorn's worker maps that through our custom `UvicornWorker` as `limit_max_requests=sys.maxsize` (recycling disabled).

Also removes the now-dead `MAX_REQUESTS` env var from the README (env table, prod runtime notes, and the sample env file).

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,7 +240,6 @@ For the **private uploads bucket**:
 | `MEDIA_ROOT` | Local media root (only used when `DEVELOPMENT=True`) | `scanning/assets/media/` |
 | `STATIC_URL` | Static file URL prefix | `static/` |
 | `NUM_WORKERS` | Gunicorn worker count | `4` |
-| `MAX_REQUESTS` | Gunicorn max requests before worker restart | `2500` |
 
 
 ### Step 2: Build the Docker Image
@@ -317,7 +316,6 @@ docker run -d \
 This starts Gunicorn with Uvicorn workers (ASGI). Configuration:
 - **Workers**: `NUM_WORKERS` env var (default: 4)
 - **Timeout**: 180 seconds
-- **Max requests**: `MAX_REQUESTS` env var (default: 2500, with 100 jitter)
 - **Bind**: `0.0.0.0:8000`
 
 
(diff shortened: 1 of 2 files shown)


grossir (approved):
LGTM!

PR #134, opened 2026-07-15
Bump blackletter to 0.1.1 and bake YOLO weights at image build

### LINKED ISSUE #131: Update Dockerfiles for next blackletter release: small/medium weights no longer bundled
Once the next blackletter version is released (freelawproject/blackletter#66, closing freelawproject/blackletter#59), the package will no longer bundle `small.pt` and `medium.pt` — all three YOLO weights are downloaded on demand from Hugging Face via `ensure_weights()`. Both Dockerfiles rely on the old bundling behavior and need updates when we bump the blackletter pin.

## `docker/django/Dockerfile`

- **`deps-web` stage:** the `rm -f /opt/venv/lib/python*/site-packages/blackletter/weights/*.pt` after `uv sync` becomes a no-op since the wheel ships no `.pt` files. Remove the line and the stale comment above it ("blackletter bundles the small/medium YOLO weights (~72 MB) as package data"). The lean web image gets the size reduction for free.
- **`deps-ml` stage (daemon / offline-local detection):** this image currently gets `small.pt`/`medium.pt` implicitly from the wheel; after the bump it will contain **zero** weights. This is not just cold-start latency — it breaks at runtime: `prod-ml` runs as `USER www-data` while `/opt/venv` is root-owned, and `ensure_weights()` downloads into `site-packages/blackletter/weights/`, so a first-use download raises `PermissionError`. It also silently breaks the offline/no-network use case this image exists for. Fix: bake the weights the daemon needs at build time in `deps-ml`, e.g. `RUN python -c "from blackletter.api import ensure_weights; ensure_weights(['small', 'medium'])"` (with the retry loop pattern from the RunPod Dockerfile if we want resilience to flaky HF).

## `scanning/runpod/Dockerfile`

- The header promises "Weights (YOLO small/medium/large + PaddleOCR PP-OCRv5) are baked in so cold start skips the network", but the warmup only runs `ensure_weights(['large'])` — the comment says it "resolves bundled small/medium and downloads large". After the bump, small/medium would silently stop being baked and would download from HF on every cold start instead (works, since this image runs as root and has `HF_HOME` set, but defeats the baked-weights goal). Fix: change the warmup to `ensure_weights(['small', 'medium', 'large'])` (or `ensure_weights()` with no args, which defaults to all three), update the comment, and extend the size-check/smoke-load assertion to cover all three files.

## Notes

- All three weights now come from a single consolidated repo: https://huggingface.co/freelawproject/blackletter-weights (`large.pt` moved there from `flooie/blackletter-large`; checksums verified identical).
- `ensure_weights()` behavior is unchanged: it resolves weights already on disk without touching the network, so baked images make it a noop at runtime.
- Blocked on the next blackletter release; do this together with the version bump in `pyproject.toml`/`uv.lock`.

quevon24:
Fixes #131

blackletter 0.1.1 no longer bundles `small.pt`/`medium.pt` in the wheel; all three YOLO weights are downloaded on demand from the consolidated HF repo `freelawproject/blackletter-weights` via `ensure_weights()`. This bumps the pin in both projects and updates everything that relied on the old bundling behavior.

## `docker/django/Dockerfile`

- **deps-web:** removed the now no-op `rm -f .../blackletter/weights/*.pt` and its stale comment. The lean web image gets the size reduction for free.
- **deps-ml:** bakes `small.pt`/`medium.pt` at build time via `ensure_weights(['small', 'medium'])` with the retry loop pattern from the RunPod Dockerfile, plus a size assertion so a silently failed download fails the build instead of shipping a weightless image. Required because `prod-ml` runs as `www-data` with a root-owned `/opt/venv` (a first-use download would raise `PermissionError`) and the offline daemon use case needs the weights on disk anyway.

## `scanning/runpod/Dockerfile`

- Warmup now runs `ensure_weights(['small', 'medium', 'large'])` so all three weights stay baked and cold start stays offline.
- The size-check/smoke-load assertion now covers all three files (each loaded with `YOLO(...)` to catch truncation).

## Version bumps

- Root `pyproject.toml`: `blackletter>=0.1.1` for both the base dep and the `local-ml` extra; `uv.lock` re-locked.
- `scanning/runpod/pyproject.toml`: `blackletter>=0.0.9` to `>=0.1.1`; its `uv.lock` re-locked (needed since the image installs with `uv sync --frozen`).
- The re-lock drops `python-doctr`, `onnx`, `ml-dtypes`, `h5py`, etc. from both lockfiles (~200 MB image-size reduction). This is the intended downstream effect of freelawproject/blackletter#65, which moved docTR into a new `[refine]` extra. We deliberately do not add `[refine]`: the only rect computation in this repo is `compute_redaction_rects(..., skip_doctr=True)` in `scanning/services.py`, and the GPU worker never computes rects. If the docTR path is ever invoked without the extra, blackletter now raises a descriptive `ImportError` rather than degrading silently.

## Related cleanups

- `scanning/runpod/handler.py`: removed the defensive `ensure_weights([... 'large'])` call in `_action_detect`, which encoded the old "small/medium are bundled" assumption. As of 0.1.1, `detect()` ensures every requested weight itself instead of silently skipping missing models, so the call was redundant.
- `scanning/runpod/README.md`: updated the stale git-pin note, the bundled-weights bullets, and the `flooie/blackletter-large` references to the consolidated repo.

## Testing

- Full suite passes against blackletter 0.1.1: 289 tests, OK (1 skipped), including the tests that exercise `detect()` and `compute_redaction_rects()` end to end.
- Both lockfiles verified with `uv lock --check`.
- The `python -c` one-liners in both Dockerfiles were syntax-checked after shell line-continuation collapse, and the size-assertion logic was exercised against fixture files for both the pass and fail cases.

### DIFF
--- a/docker/django/Dockerfile
+++ b/docker/django/Dockerfile
@@ -52,13 +52,10 @@ COPY pyproject.toml uv.lock ./
 # Base deps only. The only opt-in extras are `dev` and `local-ml`; plain sync
 # excludes both, so no torch/paddle/ultralytics lands here. Detection is
 # offloaded to RunPod, so the web tier never runs a model in-process.
-#
-# blackletter bundles the small/medium YOLO weights (~72 MB) as package data.
-# The web tier never runs detection, so drop them here — they stay in the
-# daemon image (deps-ml keeps them for local/offline detection).
+# blackletter >=0.1.1 ships no YOLO weights in the wheel, so nothing to
+# strip here; the daemon image (deps-ml) bakes the weights it needs.
 RUN --mount=type=cache,target=/root/.cache/uv \
-    uv sync && \
-    rm -f /opt/venv/lib/python*/site-packages/blackletter/weights/*.pt
+    uv sync
 
 FROM build-base AS deps-ml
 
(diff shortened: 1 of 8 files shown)


grossir (approved):
LGTM

PR #135, opened 2026-07-21
Make upload size limit configurable, default 3 GB

quevon24:
I tried to upload a 2.0 GiB scan (2,172,968,846 bytes) and it was rejected: the file was ~25 MB over the old hard-coded 2 GiB ceiling. This raises the limit to 3 GB and makes it configurable so we can adjust it without a code change next time.

## Changes

- `MAX_ORIGINAL_UPLOAD_SIZE` now lives in `settings.project.processing_storage`, sourced from a new `MAX_UPLOAD_SIZE_GB` env var (default 3 GB), instead of being hard-coded in `views.py`.
- The presign view reads the value from settings and reports the configured limit in its error message (no longer a hard-coded "2 GB" string).
- Bumped `PENDING_UPLOAD_TTL_HOURS` default from 6 to 9. That TTL must stay above the worst-case transfer time so the cleanup sweep never deletes a still-uploading pending row. A 3 GB file on a slow (~0.8 Mbps) uplink needs about 9 hours, up from ~6 hours for 2 GB.
- Updated the stale "2 GB" references in docstrings, the s3_sync helper, and the queue_detail template.

## Both enforcement layers still hold

- Fast client-facing pre-check in `presign_scan_upload`.
- Authoritative `content-length-range` condition baked into the presigned POST policy, so S3 itself rejects anything larger before it lands.

## Testing

- Updated the oversized-upload test to use 4 GB (3 GB is now a valid size).
- `scanning.tests.test_views`, `test_cleanup_command`, and `test_recover_pending_uploads` all pass.

### DIFF
--- a/scanning/settings/project/processing_storage.py
+++ b/scanning/settings/project/processing_storage.py
@@ -15,13 +15,21 @@ PROCESSING_TMP_CLEANUP_INTERVAL_SECONDS = env.int(
     "PROCESSING_TMP_CLEANUP_INTERVAL_SECONDS", default=900
 )
 
+# Maximum accepted size (in bytes) for a direct-to-S3 original PDF upload.
+# Enforced in two places: the presign view pre-checks it for a fast
+# client-facing error, and the presigned POST policy's content-length-range
+# condition lets S3 itself reject anything larger before it lands.
+# Configured in whole GB via MAX_UPLOAD_SIZE_GB (default 3 GB).
+MAX_ORIGINAL_UPLOAD_SIZE = env.int("MAX_UPLOAD_SIZE_GB", default=3) * 1024**3
+
 # How long a presigned direct-to-S3 upload (PendingUpload) may sit
 # unconfirmed before cleanup_processing_tmp deletes it -- and, if the
 # upload never landed, its fileless scan. Covers users who close the tab
 # mid-upload. Deliberately longer than S3_UPLOAD_PRESIGNED_TTL: that TTL
 # only bounds when the browser may *start* the POST (which happens
 # seconds after presign); S3 lets an in-flight upload run past policy
-# expiry, so the real bound is transfer time -- a 2 GB file on a slow
-# uplink (~0.8 Mbps) needs about 6 hours. Sweeping sooner would delete
-# the pending row and fileless scan out from under a live upload.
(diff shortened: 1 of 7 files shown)


scanning/settings/project/processing_storage.py
```diff
@@ -15,13 +15,21 @@
...
 # seconds after presign); S3 lets an in-flight upload run past policy
-# expiry, so the real bound is transfer time -- a 2 GB file on a slow
-# uplink (~0.8 Mbps) needs about 6 hours. Sweeping sooner would delete
-# the pending row and fileless scan out from under a live upload.
-PENDING_UPLOAD_TTL_HOURS = env.float("PENDING_UPLOAD_TTL_HOURS", default=6.0)
+# expiry, so the real bound is transfer time -- a 3 GB file (the
+# MAX_UPLOAD_SIZE_GB ceiling) on a slow uplink (~0.8 Mbps) needs about
+# 9 hours. Sweeping sooner would delete the pending row and fileless scan
+# out from under a live upload.
+PENDING_UPLOAD_TTL_HOURS = env.float("PENDING_UPLOAD_TTL_HOURS", default=9.0)
```
grossir, line 35: need to delete the default from this file
https://github.com/freelawproject/scanning/blob/ad648ab0b345881a2e1f683336158d73b6c998d7/scanning/management/commands/cleanup_processing_tmp.py#L128

scanning/views.py
```diff
@@ -45,8 +45,9 @@
 # Max size for a direct-to-S3 original upload. Enforced by the presigned
 # POST policy (see s3_sync.generate_presigned_post) so S3 rejects anything
 # larger before it lands, and pre-checked in the presign view for a fast
-# client-facing error.
-MAX_ORIGINAL_UPLOAD_SIZE = 2 * 1024**3  # 2 GB
+# client-facing error. Configurable via the MAX_UPLOAD_SIZE_GB env var
+# (default 3 GB); see settings.project.processing_storage.
+MAX_ORIGINAL_UPLOAD_SIZE = settings.MAX_ORIGINAL_UPLOAD_SIZE
```
grossir: why not do the import directly?

from settings import MAX_ORIGINAL_UPLOAD_SIZE

grossir (commented):
Looks good; some minor comments below

grossir (approved):
LGTM

PR #136, opened 2026-07-22
Persist RunPod job id and reattach to in-flight jobs after a daemon interruption

### LINKED ISSUE #127: Persist RunPod job id + resume in-flight jobs after a daemon interruption
## Background

When the scanning daemon is killed (SIGTERM on deploy/eviction, SIGKILL/OOM) or a
scan times out mid-pipeline, the in-flight scan is re-queued (`PROCESSING → QUEUED`)
and the whole pipeline re-runs from the affected stage on a later tick.

PR #125 (issue #124) bounds that re-queue loop and surfaces chronic offenders
(`interruption_count` + `ERROR_INTERRUPTED`). That is a **safety net** — it caps and
makes visible the wasted work, but it does not avoid it.

This issue is the **efficiency fix**: the GPU work itself is thrown away and redone on
every interruption, even when the RunPod job already finished. The daemon spends almost
all of a scan's wall-clock blocked in `runpod_client._poll` waiting on the GPU steps
(`detect` / `analyze`), so a random interruption overwhelmingly lands *while a RunPod
job is in flight* — exactly the window we currently discard.

### Two concrete costs today

1. **Redone GPU work.** On re-queue we submit a brand-new `/run` job and re-do detect/OCR
   from scratch, even if the previous job had already completed.
2. **Orphaned, double-billed jobs.** The `job_id` lives only in a local variable inside
   `runpod_client._invoke`/`_poll`. When the daemon dies it's lost, so:
   - the old job is **never cancelled** (only the timeout path calls `POST /cancel/{id}`;
     the signal handler has no id to cancel), and
   - the re-queue submits a **second** job — so both can run and bill concurrently.

The `job_id` isn't persisted **anywhere durable** — not on the `Scan`, not in Sentry
(confirmed: SIGTERM issue SCANNING-1F carries only Django SQL breadcrumbs; the
`runpod … job X submitted` logs never reach Sentry). It exists only in daemon process
memory and raw pod stdout. That absence is precisely why recovery is impossible today.

### RunPod retention makes resume feasible (and time-boxed)

The daemon uses the **async `/run`** path. Per RunPod docs:

| Item | Window |
|---|---|
| Async `/run` results via `/status/{id}` | **30 minutes** after completion, then permanently deleted |
| Job TTL (submission → deletion, any state) | **24 hours** default (queue time counts) |

A deploy/eviction restart is back in seconds, so in practice a resume attempt almost
always lands inside both windows. The edge case (outage > 30 min after the job
completed) falls back to redo — which is exactly what #125's cap is for.

### 1. Persist the job handle on the `Scan`

Add fields (names TBD):

- `runpod_job_id` (`CharField`, blank) — the current in-flight job id.
- `runpod_job_action` (`CharField`, blank) — `"detect"` or `"analyze"`, so resume knows
  which step the id belongs to (a scan runs both).
- `runpod_job_submitted_at` (`DateTimeField`, null) — submit timestamp, used to decide
  whether the result/TTL window is still open before we bother polling.

Set them right after `_submit` returns an id; **clear them** once the result is consumed
(step succeeded) or the job is known dead (FAILED/expired). Persist from
`runpod_client` (it has the id) via a small callback/return so `services` writes it to
the row.

### 2. Resume instead of resubmit

Before submitting a new job in `runpod_client._invoke` (or a new `_resume_or_submit`
helper), if the scan has a persisted `runpod_job_id` for this `action` and it's within
the TTL window, `GET /status/{job_id}` and branch:

- **`COMPLETED`** (within the 30-min result window) → use the returned `output`,
  **skip the GPU step entirely**. This is the main win.
- **`IN_QUEUE` / `IN_PROGRESS`** (within 24h TTL) → **reattach**: keep polling the
  existing job instead of submitting a duplicate.
- **`FAILED` / `CANCELLED` / `TIMED_OUT` / 404 / past the window** → clear the stale id
  and submit fresh.

### 3. Yes — this belongs on the daemon path

The re-queue → next-tick flow already re-enters the pipeline through the daemon:
`run_daemon` → `process_next_scan` → `services` (`_dispatch_analyze`, and the detect
dispatch around the `runpod_client.detect` call). So the resume check lives in
`runpod_client` and is **exercised by the daemon automatically** when it re-runs
`detect`/`analyze` for the re-queued scan — no separate worker needed. Points to cover:

- **Both interruption paths feed it:** the SIGTERM handler in `run_daemon._handle_signal`
  and the stale-recovery path in `process_next_scan._recover_stale` both just re-queue;
  resume then happens on the next tick when the stage re-runs.
- **Optional: cancel-on-shutdown.** Now that the id is persisted, `_handle_signal` *could*
  `POST /cancel/{id}` for jobs it's abandoning. Recommendation: **prefer reattach over
  cancel** — a job that's seconds from done shouldn't be killed. Cancel only as a
  fallback if reattach turns out unreliable.

## Open questions / design notes

- **Storage shape:** three fields on `Scan` vs. a small related model. Fields are simpler
  and match the existing `retry_count` / `interruption_count` pattern; lean that way
  unless we expect to track multiple concurrent jobs per scan.
- **Concurrency / multiple replicas:** the existing `status=PROCESSING` guard already
  prevents two workers claiming the same scan; resume inherits that. Confirm no
  double-poll if two replicas ever race.
- **Interaction with #125:** should a *successful* resume still bump `interruption_count`?
  The interruption did happen, but we recovered the work — arguably don't count it toward
  the `ERROR_INTERRUPTED` cap. Decide explicitly.
- **Idempotency:** resume must be per-stage. Re-running `detect` overwrites
  `detections.json`; re-running `analyze` overwrites `ocr_results`. Confirm using a
  recovered result is equivalent to a fresh run at that stage.
- **Window check is an optimization:** skip the `/status` call when
  `now - runpod_job_submitted_at` already exceeds the 24h TTL (id is guaranteed gone).

## Acceptance criteria

- [ ] `Scan` persists `runpod_job_id`, `runpod_job_action`, `runpod_job_submitted_at`;
      set on submit, cleared on consume/death (+ migration).
- [ ] On re-entry, a `COMPLETED` job's result is reused and the GPU step is skipped.
- [ ] An `IN_PROGRESS`/`IN_QUEUE` job is reattached, not duplicated.
- [ ] Expired/failed/absent ids fall back to a fresh submit cleanly.
- [ ] Resume works through the daemon (`process_next_scan`) on both the SIGTERM re-queue
      and the stale-recovery re-queue paths.
- [ ] Decision recorded on cancel-on-shutdown and on whether resume counts toward #125's
      `interruption_count`.

## Testing

- Unit: `_resume_or_submit` for each `/status` state (COMPLETED / IN_PROGRESS / IN_QUEUE /
  FAILED / 404 / past-window), with `requests` mocked.
- Unit: persisted id is set on submit and cleared on success and on dead-job fallback.
- Integration: a `PROCESSING` scan with a persisted `COMPLETED` job, run through a daemon
  tick, finishes **without** a new `/run` submission.
- Integration: an `IN_PROGRESS` job reattaches (single job id, no second `/run`).

## Out of scope

- Reducing how often the daemon is killed (pod scheduling/resources) — separate churn work.
- The interruption cap itself (PR #125 / issue #124) — this builds on top of it.

## References

- PR #125 — bounds the re-queue loop (`interruption_count` / `ERROR_INTERRUPTED`)
- Issue #124 — unbounded daemon re-queue
- Sentry SCANNING-1F — daemon SIGTERM (~510 occurrences)
- `scanning/runpod_client.py` — `_invoke` / `_submit` / `_poll` (job id currently in-memory only)
- `scanning/services.py` — `_dispatch_analyze` and the detect dispatch (GPU call sites)
- RunPod docs: [endpoint configurations](https://docs.runpod.io/serverless/endpoints/endpoint-configurations), [job states](https://docs.runpod.io/serverless/references/job-states)

quevon24:
Closes #127.

## Problem

When the daemon is killed mid-pipeline (SIGTERM on deploy/eviction, SIGKILL/OOM) or a scan times out, the in-flight scan is re-queued and the pipeline re-runs. The RunPod `job_id` lived only as a local variable inside `runpod_client._submit`/`_poll`, so on restart the daemon couldn't tell a job was already running. Two costs, both seen live on scan 1940 (2026-07-20): the killed daemon never cancelled the in-flight job (only the timeout path did), so it ran orphaned for 57m until cancelled by hand, while the re-queue submitted a duplicate job that ran concurrently; and detect/OCR was redone from scratch even when the previous job had finished.

## Fix

Persist the RunPod job handle on the `Scan` and, on re-entry, reattach to the existing job instead of submitting a duplicate.

- **Persist + reattach.** New `runpod_job_id` / `runpod_job_action` / `runpod_job_submitted_at` fields. `_invoke` records the id after submit and clears it on any terminal outcome; on a hard kill it stays set, so the next tick reattaches: a COMPLETED job's result is reused (GPU step skipped), an IN_QUEUE/IN_PROGRESS job is picked up without duplicating. A job aged out of RunPod's retention window returns 404, which the existing `_poll` path turns into a re-queue that submits fresh, so recovery is always safe.
- **Cancel orphaned jobs on shutdown.** Scans that the shutdown flags `ERROR_INTERRUPTED` are terminal and will never reattach, so their jobs are cancelled to stop billing and their persisted handle is cleared (so a later re-queue or a frontend re-detect/revalidate submits fresh rather than reattaching a dead job). Re-queued scans keep their handle so they still reattach.

## Follow-ons in this PR

- **Page-aware request timeout.** The wall-clock ceiling now scales with page count (`RUNPOD_REQUEST_TIMEOUT + RUNPOD_REQUEST_TIMEOUT_PER_PAGE * page_count`), so a large volume (e.g. 1300 pages x 3 detect models) no longer false-times-out against a flat base. A genuine timeout stays terminal with no auto-retry.
- **Aligned stale cutoff.** `_recover_stale` uses a per-scan threshold of `max(DAEMON_PROCESSING_TIMEOUT, 2x the page-aware ceiling)` instead of a single global timeout, so raising the request timeout can't leave stale recovery cutting off a large scan whose job is still legitimately running.
- **Skip re-detection on resume.** `_run_yolo` pushes `detections.json` to S3 the moment detection finishes, and `run_full_pipeline` skips detection when it already exists (restored from S3 on a resume), mirroring the existing OCR-PDF guard. So an interruption during a later stage no longer redoes detection.
- **Frontend visibility.** An error banner on the scan list surfaces ERROR scans (likely timeouts/failures), linking superusers to the admin changelist filtered to ERROR for review and re-queue, alongside the existing retry-cap and interrupted banners.

## Notes on behavior

- The biggest win (skip completed GPU work) is realized for graceful SIGTERM shutdowns, which re-queue immediately. Hard kills (SIGKILL/OOM) wait for stale recovery, by which time a COMPLETED result may have aged out (30-min retention), so they mostly get the "don't duplicate a running job" win.
- Deliberate regenerate actions (re-detect, revalidate, reprocess) are unaffected: they run through separate functions that submit unconditionally, and reviewable scans have their handle cleared, so they always submit fresh.
- A successful reattach does not un-bump `interruption_count`; the interruption still happened and the cap is about daemon churn, not job outcome.

## Testing

`DEVELOPMENT=True DB_HOST=localhost DB_SSL_MODE=prefer python manage.py test scanning.tests` — 353 tests pass. New coverage: persist/clear/reattach across every `/status` state, `cancel_job`, page-aware timeout, aligned stale cutoff, detect stage-skip and S3 persist, cancel-only-flagged on shutdown with handle clearing, and the error banner.


grossir:
This PR is now stale due to the ExternalJob model, that persists the "external_id" . Some parts may be useful and picked up by others PRs

PR #137, opened 2026-07-22
Standardize timed per-stage logging in the processing pipeline

quevon24:
## What

Adds consistent, timed INFO-level logging to every heavy stage of the scan processing pipeline, so prod logs show how long each stage took.

Previously only Tesseract OCR printed its own timing. Now bitonal conversion, OCR, YOLO detection, PaddleOCR validation, opinion pairing, and redaction-rect computation all emit the same shape:

```
Bitonal conversion...
[progress] unit='page' desc='OCR' total=943
Bitonal conversion done (137.4s, 65.2 MB)
```

## How

Introduces a `_log_stage(label, detail="")` context manager in `scanning/services.py` that logs `"<label> <detail>..."` on entry and `"<label> done (<n>s)"` on exit, using `time.monotonic()` for the elapsed measurement. Each heavy stage function is wrapped with it. The "done" line is skipped if the stage raises, so elapsed time is only reported for stages that finished.

A yielded handle lets a stage append extra context to its completion line; `_ensure_bitonal` uses it to report the output PDF size, formatted with the same `st_size / 1024 / 1024` formula blackletter uses so the numbers agree.

Also folds a few stray diagnostic `print()` calls into the logger for consistency (combined-redactions summary, the validate banner, the progress-bar line, and the non-transient pipeline-error path).

## Behavior notes

- These stage messages now go to **stderr** (via the `scanning` logger) rather than stdout, where the old `print(..., flush=True)` calls went. Docker/systemd capture both streams together, so this is not expected to matter in practice.
- Non-transient pipeline failures now route through `logger.exception` instead of `traceback.print_exc()` + `print`, which means they now **surface in Sentry** (matching the existing transient-max-retries branch). The traceback is still captured.

## Testing

Full suite passes: 330 passed, 1 skipped.

### DIFF
--- a/scanning/services.py
+++ b/scanning/services.py
@@ -5,11 +5,13 @@ daemon process.  They must NOT import Django HTTP machinery
 (HttpResponse, render, redirect, etc.).
 """
 
+import contextlib
 import json
 import logging
 import os
 import re
 import shutil
+import time
 import traceback
 from collections import Counter
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


grossir (approved):
LGTM

PR #138, opened 2026-07-22
Log RunPod submit error body and retry paused endpoints

quevon24:
## Problem

A production scan failed at YOLO detection with a 409 from RunPod's `/run` endpoint, retried 3 times, then went to ERROR. Two things made this hard to diagnose:

1. The daemon logged only the `requests` exception string (`409 Client Error: Conflict for url ...`). `_submit` calls `raise_for_status()` and never reads `r.text`, so the response body that explains the conflict was discarded. The actual cause was only recoverable by manually curl-ing the endpoint:

   ```json
   {"status":409,"title":"Conflict","detail":"Endpoint is paused (max_workers=0). Set max_workers > 0 to accept work.","code":"ENDPOINT_PAUSED"}
   ```

   The endpoint had been scaled to zero workers.

2. A submit failure raises plain `RunpodError`, which is terminal, so the scan went straight to ERROR rather than re-queuing.

## Changes

- **Log the response body on submit failure.** New `_submit_error_detail()` pulls `(status_code, code, body_text)` off the failed request; `_submit` now logs `HTTP <status> body: <text>` on each failed attempt. Transport-level errors (no response attached) log as before.
- **Classify 409 / `ENDPOINT_PAUSED` as transient.** These now raise `RunpodTransientError`, so a scaled-to-zero (or briefly unavailable) endpoint re-queues the scan instead of failing it.
- **Short-circuit the in-call retry budget on a 409.** A paused endpoint won't start accepting work within the 1-2s backoff window, so `_submit` raises transient immediately on a 409 instead of burning all `RUNPOD_MAX_RETRIES + 1` attempts. In-call retries are still used for transient transport errors (connection reset, timeout, 5xx).

## Notes

- **Bounded re-queue, not indefinite.** A persistently paused endpoint re-queues up to `RUNPOD_MAX_TRANSIENT_RETRIES` daemon ticks (default 5), then lands in `ERROR_MAX_RETRIES` with a clear message. With the short-circuit, each tick now makes a single `/run` attempt (plus the S3 presign) rather than three, so the extra traffic over the old fail-fast behavior is small and capped.
- `ENDPOINT_PAUSED` won't clear on its own (it needs `max_workers` raised); the transient handling buys a self-heal window for briefly-unavailable endpoints without permanently failing on a short blip.
- The Sentry visibility for terminal pipeline failures is handled separately in #137 (the `logger.exception` change), not here.

## Testing

Full suite: 335 passed, 1 skipped. New `_submit` tests cover: paused-endpoint-is-transient, any-409-is-transient (no `code` needed), short-circuit-does-not-retry, non-409-stays-terminal, transport-error-stays-terminal, and body-is-logged.

### DIFF
--- a/scanning/runpod_client.py
+++ b/scanning/runpod_client.py
@@ -462,37 +462,40 @@ def _submit(
             return job_id
         except Exception as exc:
             last_exc = exc
+            status_code, err_code, text = _submit_error_detail(exc)
+            body_note = (
+                f" (HTTP {status_code} body: {text[:500]})" if text else ""
+            )
+
+            # A paused / 409 endpoint won't start accepting work within
+            # the in-call backoff window, so don't burn the remaining
+            # attempts: raise transient immediately and let the daemon
+            # re-queue the scan on a later tick.
+            if status_code == 409 or err_code in _TRANSIENT_SUBMIT_CODES:
+                logger.warning(
+                    "runpod submit failed: %s%s; endpoint not accepting "
+                    "work, re-queueing",
+                    exc,
+                    body_note,
+                )
+                raise RunpodTransientError(
+                    f"RunPod endpoint not accepting work (HTTP "
+                    f"{status_code}): {exc}"
+                ) from exc
+
             sleep_for = 2**attempt
-            status_code, _, text = _submit_error_detail(exc)
-            if text:
-                logger.warning(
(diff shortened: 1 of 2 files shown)


grossir (approved):
LGTM!

PR #139, opened 2026-07-23
Clarify admin re-queue actions for errored scans

quevon24:
## Problem

An operator looking at a scan in `Error` status in the Django admin couldn't tell which of the three re-queue actions to pick. The action names didn't line up with the status labels, and the two status-scoped actions (`requeue_retry_cap_scans`, `requeue_interrupted_scans`) silently reported "Re-queued 0 scan(s)" as a **success** when run on a plain `Error` scan, so it looked like nothing happened for no reason.

There are three distinct error statuses (`Error`, `Error (retry cap hit)`, `Error (interrupted too often)`), plus a scan can be stuck in `Processing` after a killed daemon.

## Changes

- **New general action: "Re-queue selected scans (any Error / stuck Processing)".** Replaces the old `reset_to_queued`. Requeues a scan in any error status or stuck `Processing`, and resets both `retry_count` and `interruption_count`. This is the obvious pick for a plain `Error` scan. Unlike the old action it filters to error/processing statuses, so a stray selection (a `Pending Review` or `Approved` scan) is skipped and reported rather than silently reset.
- **Scoped actions relabeled to name their exact status** ("retry cap hit only", "interrupted too often only") so it's clear what each targets and what counter it resets.
- **No more misleading silent success.** All three actions now warn (instead of reporting a clean success) when the selection contained no matching scan, or when only part of the selection matched, and they report how many were skipped/left untouched. The scoped actions stay surgical: they only ever touch their one status.

## Behavior notes

- The general action treats `Processing` as re-queueable, so it will reset an actively-processing scan, not just a stuck one (the admin can't distinguish them). This matches, and is safer than, the old unfiltered `reset_to_queued`.
- No DB schema change; actions only touch `status`, the two counters, and the `progress_*` fields.

## Testing

Full suite: 342 passed, 1 skipped. New `scanning/tests/test_admin.py` covers the general action (plain error, all error/processing statuses, dual-counter reset, progress-field reset, skip-and-report, empty selection, warn-on-no-match) and the scoped actions (match, warn-on-plain-error, and mixed-selection leaving non-matching scans fully untouched).

### DIFF
--- a/scanning/admin.py
+++ b/scanning/admin.py
@@ -134,7 +134,7 @@ class ScanAdmin(admin.ModelAdmin):
     ]
     date_hierarchy = "date_created"
     actions = [
-        "reset_to_queued",
+        "requeue_scans",
         "requeue_retry_cap_scans",
         "requeue_interrupted_scans",
     ]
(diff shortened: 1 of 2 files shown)


grossir (approved):
This looks good to merge

I think you could improve this PR a bit by abstracting out repeated logic. All follow the same pattern, you would just need to pass the different templates for each case
```
if not updated:
     message
     return

skipped computation
if skipped:
     message addition

message user
```

grossir (approved):

Issue #141, opened 2026-07-25
Audit what is left in checker.css


quevon24:
`scanning/static/scanning/checker.css` predates the Tailwind migration and is loaded only by `scan_process.html`, after the Tailwind bundle. Its selectors are bare elements or single classes, so they tie with Tailwind utilities on specificity and win on source order. Utilities on the element lose to rules in this file, silently.

#140 fixed the dark mode contrast that came out of that and removed the worst offenders, the global `body` colors and the `.issue-card` background. What is left is the audit.

Worth investigating:

- How much of the file is still used at all. Blocks like the navbar, document cards, report layout, issues panel, OCR summary and stepper look like leftovers from before the migration.
- Which of the remaining rules is genuinely viewer specific chrome that belongs in a plain stylesheet, and which should move to a Tailwind component class in `scanning/assets/tailwind/input.css` alongside `.btn-primary` and friends.
- `.viewer-panel` hard codes `background: #525659` in both schemes and overrides the `bg-white dark:bg-gray-900` on its element. Decide whether the viewer backdrop should stay fixed or follow the scheme, and fix its placeholder text contrast accordingly.
- Whether the file can go away entirely, taking the `<link>` in `scan_process.html` with it.

Note that deadness here is not greppable. Some class names are assembled in JS strings so they never appear in a class attribute, many rules are compound modifiers that only mean anything next to a base class, and the resize handle classes are single letters. Each rule needs checking against the element that actually carries it, in the browser, in both color schemes.

PR #142, opened 2026-07-25
fix(scanning): correct light and dark mode contrast in scan templates

### LINKED ISSUE #140: Retire checker.css from the process page
`scanning/static/scanning/checker.css` is a pre-Tailwind stylesheet loaded only by `scan_process.html`, and it loads after the Tailwind bundle. Its selectors are bare elements or single classes, so they have the same specificity as Tailwind utilities and win on source order. Utilities on the element lose to rules in this file, silently.

It is also written entirely for light mode, which is where the dark mode contrast bugs on the process page came from: a global `body` color that overrode Tailwind's scheme aware one, and an `.issue-card` background that overrode the template's dark variant. Those are fixed, but the rest of the file carries the same assumption, so the next template edit can reintroduce the problem somewhere else.

Worth investigating:

- How much of the file is still used at all. A fair amount looks like it predates the Tailwind migration (navbar, document cards, report layout, stepper).
- Which of the remaining rules is genuinely viewer specific chrome that belongs in a plain stylesheet, and which should move to a Tailwind component class in `scanning/assets/tailwind/input.css` alongside `.btn-primary` and friends.
- Whether the file can go away entirely, taking the `<link>` in `scan_process.html` with it.

Note that deadness here is not greppable. Some class names are assembled in JS strings so they never appear in a class attribute, many rules are compound modifiers that only mean anything next to a base class, and the resize handle classes are single letters. Each rule needs checking against the element that actually carries it, in the browser, in both color schemes.

quevon24:
Several elements on the process page were unreadable when the OS is in dark mode. The `Issues (N)` and `Pages (N)` sidebar headings disappeared into the sidebar, and inside an issue card the severity line and the Dismiss button looked fine while the message paragraph was invisible.

Closes #140.

## Root cause

`scanning/static/scanning/checker.css` is a pre-Tailwind stylesheet loaded only by `scan_process.html`, and `{% block head %}` puts it after the Tailwind bundle. Its selectors are bare elements or single classes, so they tie with Tailwind utilities on specificity and win on source order. Two rules did the damage:

- `body { background: #f5f5f5; color: #333 }` beat Tailwind's `dark:text-gray-100`, so every element without an explicit color class inherited dark grey. The sidebar `h2`s carry no color class, which put them at 1.4:1 against the `gray-900` sidebar. Note that the custom palette maps `gray-900` to `#101828`, which is why it reads as navy.
- `.issue-card { background: #fafafa }` beat the template's `dark:bg-blue-900/20`, so cards stayed near white while the message paragraph rendered in `dark:text-gray-300`, also 1.4:1. This is also why the severity label and Dismiss button looked correct: they resolved to `#333` and `bg-green-600 text-white` respectively. The same rule was quietly defeating `bg-red-50` / `bg-yellow-50` / `bg-blue-50` in light mode too, so severity tinting on issue cards starts working now.

After the fix: heading 16.1:1, issue message 11.0:1, severity label 14.8:1, page label 6.3:1.

## Changes

`checker.css` stops imposing a light theme:

- Dropped `background` and `color` from `body`. The Tailwind base layer already sets both per scheme, and the headings simply inherit correctly once this is out of the way.
- Dropped `.issue-card`'s `background`, `border-left`, `:hover` and its four never-applied severity variants. The scheme-aware Tailwind classes on the element now own its color. Kept the padding and margin, which currently override `px-2 py-1.5 mb-1`, so removing them would shift spacing.
- Pinned explicit text colors on `.page-container` and `.dupe-modal`, the surfaces that are white in both schemes and were relying on the inherited `#333`.
- Deleted the `.opinion-card.selected, .issue-card.selected` rule. It is unreachable: `.selected` only ever lands on `.opinion-nav` and `.opinion-card`, no `.opinion-card` element is ever created, and step 3 marks the active card with `style.outline` instead.

Templates and viewer JS:

- Replaced the uncovered-headnote chip's inline `background-color: yellow` with `bg-yellow-300 text-yellow-950`. It depended on the inherited `#333` and would otherwise have gone light on yellow.
- Added the `hover:` and `dark:hover:` pairs that replace the deleted `.issue-card:hover`, on both the step 1 and step 3 cards. The `dark:hover:` half is required, not decorative: `.hover\:bg-red-100:hover` is 0-2-0 and would beat `.dark\:bg-red-900\/20` at 0-1-0, flashing light red on hover in dark mode.
- Paired the remaining unpaired grey and primary text across `scan_process.html`, `page_detail.html`, `pages_list.html`, `queue_detail.html`, `opinion_detail.html` and `viewer_progress.js`. Two of these were failing in light mode rather than dark, and the `text-primary-600` links were at 2.1:1 in dark mode because the base layer's `dark:text-primary-400` sits on the bare `a` selector and loses to the element's own class regardless of order. Seven `hover:text-primary-600` links had no dark pairing, so hovering made them harder to read than not hovering.

## Tailwind content glob

Second commit, separable. `tailwind.config.js` listed `../static-global/js/**/*.js`, a directory that does not exist, and the five viewer JS files that build markup at runtime live in `scanning/static/scanning/` and were never scanned. Any Tailwind class used only in those files silently failed to compile.

Measured rather than assumed: built the bundle with and without the added glob and diffed the generated selectors. 10 new, 0 lost. One is `border-orange-400`, the real fix. The other nine (`contents`, `filter`, `grow`, `outline`, `resize`, `transform`, `visible`, `ease-in-out`, `!container`) are false positives where the extractor picks up ordinary JS words that happen to be utility names; no element carries any of them as a class, so they are inert, about 1.5KB. The real rebuild matched that prediction exactly at 396 selectors.

**One visible change to expect:** the DUP divider in the processing sidebar (`viewer_progress.js:58`) currently renders in `currentColor` because `border-orange-400` was missing. It will now be orange, matching its own label and the GAP and ORDER dividers that already worked.

## Testing

335 tests pass, though this is CSS and templates so that only confirms nothing else moved.

Worth a look in both schemes, since the visual check is the real one here. Load `/scans/<pk>/process/` at `?step=1`, `?step=2` and `?step=3`:

- Sidebar headings, issue cards and their hover, page rows, opinion rows, unmatched caption and key icon rows, generated file list.
- PDF viewer chrome unchanged: white page containers with dark labels on the dark backdrop, zoom toolbar, page label buttons still wrapping at narrow widths.
- The duplicate comparison modal, reachable from a `duplicate_page` issue.
- The processing sidebar, which is regenerated by `viewer_progress.js` rather than rendered from the template.

## Known side effect

Deferring `body` to the base layer changes the light mode page background from `#f5f5f5` to `gray-25` `#FCFCFD`, and in dark mode the sidebar's `dark:bg-gray-900` is now the same value as the page background. Only the border separates them, in both schemes. Happy to add a surface tint if that reads badly.

## Left alone deliberately

- The dead rules in `checker.css` (navbar, doc cards, report layout, issues panel, OCR summary, stepper). Several class names there are assembled in JS strings, so deadness needs per-rule checking rather than a grep. Tracked in #141.
- `.viewer-panel` forces `#525659` in both schemes, so its placeholder text was 2.9:1. Bumped the two placeholders to `text-gray-300`, but the panel itself is viewer chrome and belongs with #141.
- `queue.html` and `opinion_detail.html` pair `text-gray-400` with `dark:text-gray-500`, which is weak in both schemes (2.6:1 and 3.5:1). That is a deliberate looking choice about how muted secondary text should be, so it wants one decision applied consistently rather than a drive-by here.

### DIFF
--- a/scanning/assets/tailwind/tailwind.config.js
+++ b/scanning/assets/tailwind/tailwind.config.js
@@ -46,6 +46,9 @@ module.exports = {
 
       /* JS files that could contain Tailwind CSS classes */
       '../static-global/js/**/*.js',
+
+      /* App JS that builds markup at runtime (the process page viewer) */
+      '../../static/**/*.js',
     ],
   },
   theme: {
--- a/scanning/static/scanning/checker.css
+++ b/scanning/static/scanning/checker.css
@@ -1,9 +1,10 @@
 * { box-sizing: border-box; margin: 0; padding: 0; }
 
+/* Background and text color come from the Tailwind base layer so they follow
+   the light/dark scheme. Setting them here would win on source order and pin
+   the whole page to a light theme. */
 body {
     font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, sans-serif;
-    background: #f5f5f5;
-    color: #333;
 }
 
 .navbar {
@@ -248,20 +249,15 @@ a.doc-name:hover { text-decoration: underline; }
 .dismiss-btn { background: #059669; font-size: 11px; padding: 3px 12px; }
 .dismiss-btn:hover { background: #047857; }
 
+/* Layout only. Background, border and hover come from the scheme-aware
+   Tailwind classes on the element itself. */
 .issue-card {
     padding: 10px 12px;
     margin-top: 8px;
     border-radius: 6px;
     cursor: pointer;
-    border-left: 4px solid #ddd;
-    background: #fafafa;
     transition: background 0.15s;
 }
-.issue-card:hover { background: #f0f0f0; }
-.issue-card.ok { border-left-color: #22c55e; background: #f0fdf4; }
-.issue-card.error { border-left-color: #ef4444; }
-.issue-card.warning { border-left-color: #f59e0b; }
-.issue-card.info { border-left-color: #3b82f6; }
 
 .issue-header { display: flex; align-items: center; gap: 8px; }
 .severity-badge {
@@ -344,8 +340,11 @@ a.doc-name:hover { text-decoration: underline; }
 }
 
 /* Page containers */
+/* Scanned pages stay white paper in both schemes, so they carry their own
+   dark text color rather than inheriting the scheme-aware one. */
 .page-container {
     background: white;
+    color: #333;
     box-shadow: 0 2px 8px rgba(0,0,0,0.3);
     position: relative;
 }
@@ -560,6 +559,7 @@ a.doc-name:hover { text-decoration: underline; }
 }
 .dupe-modal {
     background: white;
+    color: #333;
     border-radius: 10px;
     padding: 20px;
     max-width: 92vw;
@@ -643,10 +643,6 @@ a.doc-name:hover { text-decoration: underline; }
     max-height: 400px;
     overflow-y: auto;
     font-family: monospace;
-}
-.opinion-card.selected, .issue-card.selected {
-    border-left-color: #7c3aed;
-    background: #ede9fe;
 }
 /* Detection overlay boxes */
 .detection-box {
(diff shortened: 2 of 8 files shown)


grossir (approved):
LGTM

Issue #68, opened 2026-05-12
Scanning Q/A Evaluations Criteria and datasets


flooie:
Using this to detail the different criteria being used to validate the quality of the output.

flooie:
# Quality + cost analysis across prompts and models

## Summary

I've spent a chunk of time building a ground-truth dataset and the infrastructure to measure extraction quality across prompt + model combinations. This issue documents the methodology and the open questions on which combination to ship.

## Ground-truth dataset

628 hand-curated pages, split into two corpora by use case:

| Corpus | Pages | Role |
|--------|------:|------|
| `short/` |    48 | Single-page cases — denial orders, per-curiam dispositions, short orders. Smoke tests; cheap to iterate against. |
| `long/` |   584 | Multi-page cases across 28 volumes. **Primary regression signal** — meaningful divergence here is a real defect, not noise. |

The corpora are deliberately small and curated rather than scraped. Each page was hand-verified, so a divergence between extraction and ground truth is grounds for investigation, not "probably ground truth drift."

the corpus was originally generated when we planned two differnet prompts so there some first pages were changed because the masking is no longer part of the process and ending pages should have stamps.

## Why not a single match score

Ground truth drift. They were generated by earlier prompt and model versions, so they carry OCR artifacts the current model corrects, older tag conventions, hyphenation choices that differ, and column-wrap paragraphs split differently. A byte-for-byte match score would mostly measure that drift, not real quality.

We also can't lean on text similarity alone — a model can produce *all the right words* and still be wrong (footnotes placed in `<opinion>`, missing case-end markers, hallucinated tag names, broken footnote/footnotemark linking). And it can produce *different words* while being right (joined hyphenations, corrected OCR, recovered parallel citations the ground truth missed).

The methodology has to separate "the model is wrong" from "the ground truth is stale" from "valid representation difference." That's why we measure many narrow things instead of one broad thing.

## Methodology — multi-dimensional scoring

Each page produces a `score` dict with ~20 independent dimensions, each in `[0, 1]`. They roll up into four composites. Full reference in [`tests/tests.md`](../tests/tests.md); here's the shape:

### Composites (roll-ups)

| Name | What it averages | When to use it |
|------|------------------|----------------|
| **`composite_core`** | 10 load-bearing dimensions (next table) | The headline — if it drops, investigate |
| **`composite_tags`** | `text_order`, `no_hallucinated_tags`, 7 `tag_*` count dimensions | Schema / convention drift signal |
[...]

PR #146, opened 2026-08-01
Stop embedding a Tesseract text layer in the processing pipeline

### LINKED ISSUE #145: Stop embedding a Tesseract text layer in the processing pipeline
## What happens today

`run_full_pipeline` runs ocrmypdf/Tesseract over `bitonal.pdf` as step 3, between bitonal conversion and YOLO detection (`services.py:1750`). The pass lives in `services._run_ocr` / `services._ocr` (`services.py:368` and `services.py:408`), which is a local fork of `blackletter.api.ocr` carrying a progress-bar plugin. It writes `<reporter>.<volume>.<first>.<last>.pdf`, and that file then becomes the de facto "processing PDF" for the rest of the pipeline: about ten call sites resolve it through `utils.find_ocr_pdf` for page geometry, previews, redaction rects and generation. `run_detect` re-runs the pass when the file is missing, and `_ocr_single_page` re-OCRs a single page after a smart insert.

## Why it should go

It is by far the most CPU-hungry step in the pipeline, it runs inside the daemon process, and ocrmypdf fans Tesseract out across every available core while it does. Timed on a 30-page slice of a real volume with the process pinned to a fixed number of cores:

| cores | 30 pages | per page | scaled to a 1292-page volume |
| --- | --- | --- | --- |
| 22 (dev box) | 39.3s | 1.31s | ~28 min |
| 8 (closer to the server) | 52.8s | 1.76s | ~38 min |
| 4 | 73.1s | 2.44s | ~52 min |

So on an 8-core server a full volume spends something like half an hour with every core saturated, and it does so *before* detection and page-number analysis, which is the part a reviewer is actually waiting for. Two caveats on those numbers: the pass scales poorly with cores (22 to 8 costs only 1.3x, because the serial phases such as linearize and save dominate at this sample size), and per-page cost depends on the scan, since a cleaner volume measured 15s for the same 30 pages on the same box.

Nothing in step 1 or step 2 of the review UI reads the text it produces. The page-number sidebar, the detections, the pairing and the review issues all come from YOLO plus PaddleOCR, not from the embedded layer.

## What the text layer was buying us

Word-level bounding boxes, and they were genuinely useful for geometry rather than for reading:

- Redaction adjustment. Headnote rects are tightened and their bottom and side edges clamped from the word boxes, so a block rect derived from detection geometry ends up hugging the printed text. The boxes also gave outward slack: because a word overlapping a rect pulls the bounds outward, a rect whose edge clipped a character absorbed the whole character for free.
- The page content box that the margin whiteouts are derived from (`blackletter.margins` reads `get_text("blocks")`).
- Protecting printed page numbers from being redacted, by tightening the protected rect to the actual digits.
- Blank-page detection at generate time, which extracts text from the post-redaction per-page PDFs.
- Searchable delivered opinions, since the redacted and unredacted files inherit the layer.

The first four are geometry, and geometry can be measured from the scanned pixels instead. Only the last one needs real text, and it only matters for the final opinion files.

## What we want

The pipeline should not embed a text layer at all. `bitonal.pdf` should be the processing PDF end to end, with the original PDF as the fallback, and the review steps should be unaffected.

If we still want searchable opinions, that belongs at the very end of Generate Files, over the pages that have already been redacted and masked, and it should come from blackletter as an optional step rather than from a fork here. See https://github.com/freelawproject/blackletter/issues/68.

## Not in scope

Tesseract stays a dependency and stays in the images. `blackletter.analyze` uses it as a per-crop fallback digit reader when PaddleOCR cannot read a page number, which costs milliseconds and is unrelated to this pass.

## Risks to keep in mind

Several things silently depend on the layer existing, and would need attention rather than deletion: margin rects (no text means no content box, which quietly disables margin cleanup), the blank-page check at generate time (no text means every page reads as blank), the `find_ocr_pdf(...) or scan.pdf_path` fallbacks (falling back to the multi-GB original means an S3 pull per request in prod), and the searchability of the deliverables.

### LINKED ISSUE #155: Remove Tesseract and ocrmypdf entirely: crop fallbacks, LLM text layer, and their dependencies
Follow-up to #146, interim until the #149 cutover.

## Goal

Two things at once: make PaddleOCR the only page-number reader, and get every OCR-embedding dependency out of the images. After #146 nothing embeds a text layer as a pipeline stage, but Tesseract is still installed and still runs — `blackletter.analyze` tries it on every page-number crop Paddle cannot read, and `LLM_PAGE_TEXT_LAYER` can still invoke ocrmypdf over the `llm/` pages.

When Paddle cannot read a crop, the page gets `detected=None` and surfaces as the existing `no_page_number` issue for manual assignment in review 1. That path already exists (`Issue.CheckName.NO_PAGE_NUMBER`, `models.py:895`, emitted from `analysis["not_detected"]` at `validate.py:350`), and `_auto_correct` already interpolates isolated misses from their neighbours. Nothing new is needed downstream.

## What runs today

`analyze._ocr_crop_multi` (`analyze.py:120-215`) tries three backends per crop:

1. **PaddleOCR** — what we keep
2. **Tesseract** via pytesseract, `--psm 7 digits`, recorded as `score: 0.5`, `ocr: "tesseract"`
3. **GLM-OCR** — lazy-loads `zai-org/GLM-OCR` through transformers/torch when the first two fail

## What we want

Paddle, then `None`. **Both** fallbacks go: removing only Tesseract promotes GLM, which is strictly worse — a VLM downloaded and loaded at runtime, on CPU in the `prod-ml` tier, per volume. That is the costly path this issue exists to close.

## Work

**blackletter**
- Remove steps 2 and 3 from `_ocr_crop_multi`, or gate them behind an argument defaulting off. Prefer an explicit switch over deleting `pytesseract` and letting the guarded `try/except ImportError` swallow it — a toggle that silently re-enables itself when something pulls the dependency back in transitively is not a toggle.
- **Move `ocrmypdf` out of the base dependency list into an extra** (`blackletter[ocr]` or similar). It is currently required by every install (`pyproject.toml:40`), so nothing scanning does on its own removes it from the image. Since #69 made the OCR pass optional, base no longer needs it — `api.ocr` and `api.add_text_layer` become extra-gated.
- Drop `pytesseract` from blackletter's dependencies if nothing else uses it.

**scanning**
- Remove `pytesseract>=0.3.13` and `ocrmypdf>=17` from `pyproject.toml`; install blackletter without the OCR extra.
- Remove `tesseract-ocr` / `tesseract-ocr-eng` from both Dockerfile tiers (`prod-web`, `prod-ml`).
- Delete `LLM_PAGE_TEXT_LAYER` and `services._add_llm_page_text_layer`. This is decided, not a question: with no Tesseract there is nothing for the flag to run.
- Remove the now-dead text crops in `ai/user_prompt.py` — the caption first line, the column-top continuation and the footnote band all read a text layer that no longer exists, so they return empty on every page. The structural roadmap built from the detections stays; only the crop helpers and the `pdfplumber.open` at `user_prompt.py:411` go.
- Drop `pdfplumber>=0.11` — `ai/user_prompt.py` is its only consumer.

## Measure first

`ocr_results` records which backend read each page, so the blast radius is one query over stored scans, no re-runs:

```python
from collections import Counter
from scanning.models import Scan
c = Counter()
for r in Scan.objects.exclude(ocr_results=[]).values_list("ocr_results", flat=True):
    c.update(e.get("ocr") for e in r if e.get("detected"))
print(c)
```

The `tesseract` count is how many extra `no_page_number` issues per volume reviewers would absorb. Isolated misses are interpolated away by `_auto_correct`; what matters is how often they land on consecutive pages, where interpolation cannot help.

## Risks

- **Reviewer cost lands in review 1**, the stage we are trying to make fast and pleasant (#151). If the measured count is not near zero, reconsider doing this before the cutover.
- **The LLM prompt loses its text snippets** until the ensemble replaces that stage. Accepted: the model reads the page image, and #148 rewrites this path anyway.
- **`score: 0.5`** was Tesseract's marker and `_auto_correct` consumes score, so the distribution shifts slightly. Check nothing keys off that value.

## Sequencing

#149 deletes the whole `analyze` path — Paddle, both fallbacks and all — at cutover. The page-number half of this issue happens for free then; the dependency and image half does not, and is worth doing on its own (#119).

## Acceptance criteria

- No `tesseract-ocr` in either image; no `pytesseract`, `ocrmypdf` or `pdfplumber` in the resolved dependency tree.
- Image size recorded before and after, both tiers.
- A crop Paddle cannot read produces `detected=None` and a `no_page_number` issue, with no other model loaded.
- Backend counts from the query above recorded here, before and after.
- Review 1 completes end to end on a volume with unreadable page numbers; Generate Files still produces the same deliverables.

## Related

#146 (removed the text-layer pass) · #149 (deletes this path at cutover) · #148 (parent cleanup) · #119 (image size) · #151 (review-1 UX, where the cost lands) · blackletter#69 (made the OCR pass optional, which is what lets it leave the base install)

quevon24:
Fixes #145. Depends on freelawproject/blackletter#69.

## Why

`run_full_pipeline` ran ocrmypdf/Tesseract over every page as step 3, before detection and page-number analysis. On an 8-core server a full volume spent something like half an hour with every core saturated, in the daemon process, before the reviewer saw anything.

Nothing in the review UI read the text it produced. What the layer was actually buying us was word bounding boxes, and only for geometry: tightening headnote rects, deriving the margin whiteouts, protecting printed page numbers, and deciding whether a generated page was blank. A scanned page is black marks on white, so all four can be measured from the pixels instead. That work is now in blackletter (#68), and this branch takes the pass out and consumes the library.

## What changed

**The pass is gone.** `_run_ocr`, `_ocr`, `_ocr_single_page` and the OCR'd PDF are deleted. `bitonal.pdf` is the processing PDF end to end, resolved through the new `utils.processing_pdf_path`, with the original as a fallback that in production means an S3 pull and is treated as the slow path. Scans processed before this still have their OCR'd PDF and still use it.

**Geometry comes from blackletter, not from here.** `scanning/ink.py` and `scanning/margins.py` are deleted, along with the passes this app used to run over `compute_redaction_rects`' output: column snapping, splitting at headnote detections, growth onto adjoining ink, and pulling margin strips back off real detections. They run inside the library now, so every caller gets the same rects. `_build_combined_redactions` delegates to `api.build_redactions`, the blank-body check to `process.page_body_covered`, and the page-sequence analysis to `validate.build_analysis`. About 1,300 lines of app code removed.

**Searchable LLM pages are opt-in.** `ai.user_prompt` crops three text snippets off each generated page, and with no text layer they come back empty. Rather than reinstate the pass for the whole pipeline, `LLM_PAGE_TEXT_LAYER` runs blackletter's `add_text_layer` over `llm/` at the end of Generate Files, after redaction and masking. Off by default: the model reads the page image anyway.

**One-file S3 pulls.** Three views pulled the scan's entire processing prefix to serve a single file, which is gigabytes on a full volume, and the sync views share one executor under ASGI.

## Behaviour changes worth knowing

Re-checking a scan without applying edits now agrees with the other two recalculation paths, where it used to differ in three ways: a page printed as a range accounts for every number it covers, out-of-range readings reach the issue builder, and a scan with no end page gets missing-page findings across the span it detected.

## Verification

404 tests, up from 374. Checked against a real 1,292-page volume by recomputing its geometry and diffing against what the pre-migration code had stored:

- Ink left uncovered around headnote rects, over 200 pages: **415 px → 198 px**
- Content ink erased by margin strips: **1 px, unchanged**
- Column ink no longer covered that sits inside a HEADNOTE detection: **4 px** across three pages, one to two pixels each
- Column ink no longer covered that sits outside one: 10,083 px, the court's own text the old rects were blacking out

## Before merging

`pyproject.toml` pins blackletter to the #68 branch and `uv.lock` resolves it to a commit, because the ink geometry is unreleased. Once blackletter #69 is merged and tagged, the last commit here has to raise the floor to `blackletter>=0.2.0` and delete the `[tool.uv.sources]` entry. The branch this pin names stops existing when #69 merges.

### DIFF
--- a/scanning/services.py
+++ b/scanning/services.py
@@ -18,9 +18,11 @@ from pathlib import Path
 
 import django
 import fitz
-import pdfplumber
 from blackletter.api import (
     bitonal as bl_bitonal,
+)
+from blackletter.api import (
+    build_redactions as bl_build_redactions,
 )
 from blackletter.api import (
     pair as bl_pair,
@@ -37,13 +39,18 @@ from blackletter.models import (
 from blackletter.models import (
     Document as BLDoc,
 )
-from blackletter.process import compute_redaction_rects
-from blackletter.scanner import _pair_opinions
+from blackletter.process import compute_redaction_rects, page_body_covered
+from blackletter.scanner import (
+    _pair_opinions,
+    snap_document_columns,
+    snap_text_columns_to_ink,
+)
 from blackletter.validate import (
     _auto_correct,
-    _build_issues,
-    _parse_expected_range,
     _split_in_out_of_range,
+    build_analysis,
+    build_issues,
+    parse_expected_range,
 )
 from django.conf import settings
 from django.db.models import Case, F, Value, When
@@ -60,7 +67,13 @@ from scanning.models import (
     Status,
     Volume,
 )
-from scanning.utils import ensure_output_dir, find_ocr_pdf, has_s3_credentials
+from scanning.utils import (
+    ensure_output_dir,
+    find_ocr_pdf,
+    find_processing_pdf,
+    has_s3_credentials,
+    processing_pdf_path,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -406,160 +419,6 @@ def _ensure_bitonal(scan: "Scan", output_dir: Path) -> Path:
     return bitonal_path
 
 
-def _run_ocr(
-    scan_pk: int,
-    bitonal_path: str,
-    output_dir: str,
-    reporter: str,
-    volume: str,
-    first_page: int,
-) -> str:
-    """Run Tesseract OCR via blackletter.
-
-    :param scan_pk: Primary key of the scan (for progress updates).
-    :param bitonal_path: Path to the bitonal PDF.
-    :param output_dir: Directory to write the OCR'd PDF into.
-    :param reporter: Reporter short name (e.g. "f3d").
-    :param volume: Volume number as a string.
-    :param first_page: First logical page number in the PDF.
-    :return: Path to the OCR'd PDF.
-    """
-    with fitz.open(bitonal_path) as pdf:
-        total_pages = pdf.page_count
-    _update_progress(
-        scan_pk,
-        f"Running Tesseract OCR (0/{total_pages} pages)...",
-        current=0,
-        total=total_pages,
-    )
-    ocr_path = _ocr(
-        scan_pk,
-        bitonal_path,
-        output_dir,
-        reporter=reporter,
-        volume=volume,
-        first_page=first_page,
-        total_pages=total_pages,
-    )
-    # Persist the OCR PDF as soon as it exists (see _ensure_bitonal).
-    _push_generated_file_to_s3(scan_pk, Path(ocr_path).name)
-    return ocr_path
-
-
-def _ocr(
-    scan_pk: int,
-    pdf_path: str | Path,
-    output_dir: str | Path,
-    reporter: str = "",
-    volume: str = "",
-    first_page: int = 1,
-    total_pages: int = 0,
-    language: str = "eng",
-) -> Path:
-    """OCR a PDF (add text layer via ocrmypdf/Tesseract).
-
-    Temporary local copy of blackletter's ``ocr`` function with
-    progress callback support. Will be moved to blackletter once
-    validated.
-
-    :param scan_pk: Primary key of the scan (for progress updates).
-    :param pdf_path: Path to the input PDF.
-    :param output_dir: Directory to write the OCR'd PDF into.
-    :param reporter: Reporter short name (e.g. "f3d").
-    :param volume: Volume number as a string.
-    :param first_page: First logical page number in the PDF.
-    :param total_pages: Total page count (for progress messages).
-    :param language: Tesseract language code.
-    :return: Path to the OCR'd PDF.
-    """
-    import logging
-
-    import ocrmypdf
-    from ocrmypdf import hookimpl
-    from ocrmypdf._plugin_manager import get_plugin_manager
-
-    pdf_path = Path(pdf_path)
-    output_dir = Path(output_dir)
-
-    # Build output filename
-    last_page = first_page + total_pages - 1
-    parts = [
-        p
-        for p in [reporter, str(volume), str(first_page), str(last_page)]
-        if p
-    ]
-    scan_name = ".".join(parts) if parts else pdf_path.stem
-    output_path = output_dir / f"{scan_name}.pdf"
-
-    # Suppress noisy loggers
-    for name in (
-        "pikepdf",
-        "fontTools",
-        "fontTools.subset",
-        "fontTools.ttLib",
-        "ocrmypdf",
-    ):
-        logging.getLogger(name).setLevel(logging.ERROR)
-    for _n in list(logging.root.manager.loggerDict):
-        if _n.startswith("ocrmypdf"):
-            logging.getLogger(_n).setLevel(logging.ERROR)
-
-    # Progress bar class that updates the scan record per page
-    class _ScanProgressBar:
-        def __init__(
-            self, *, total=None, desc=None, unit=None, disable=False, **kw
-        ):
-            self._total = total or total_pages
-            self._unit = unit
-            self._desc = desc
-            self._current = 0
-            logger.info(
-                "[progress] unit=%r desc=%r total=%s", unit, desc, total
-            )
-
-        def __enter__(self):
-            return self
-
-        def __exit__(self, *a):
-            return False
-
-        def update(self, n=1, *, completed=None):
-            self._current += n
-            if self._unit == "page" and (
-                self._current % 10 == 0 or self._current == self._total
-            ):
-                _update_progress(
-                    scan_pk,
-                    f"Tesseract OCR: {self._current}/{self._total} pages...",
-                    current=self._current,
-                    total=self._total,
-                )
-
-    class _ScanProgressPlugin:
-        @hookimpl
-        def get_progressbar_class(self):
-            return _ScanProgressBar
-
-    pm = get_plugin_manager()
-    pm._pm.register(_ScanProgressPlugin())
-
-    with _log_stage(
-        "OCR", f"{total_pages} pages ({os.cpu_count() or 1} CPUs)"
-    ):
-        ocrmypdf.ocr(
-            str(pdf_path),
-            str(output_path),
-            pdf_renderer="auto",
-            optimize=1,
-            output_type="pdf",
-            language=[language],
-            tesseract_timeout=120,
-            progress_bar=True,
-            plugin_manager=pm,
-        )
-    return output_path
-
-
 def _run_yolo(scan_pk: int, pdf_path: str, output_dir: str) -> None:
     """Run YOLO detection via runpod_client (local or remote).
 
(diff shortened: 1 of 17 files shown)


scanning/services.py
```diff
@@ -641,6 +500,67 @@ def _remote_progress(current, total, message):
...
+                    bbox=BBox(x1=d.x0, y1=d.y0, x2=d.x1, y2=d.y1),
+                    label=Label.TEXT_COLUMN,
+                    confidence=d.confidence,
+                    page_index=page_index,
+                )
+                for d in columns
+            ]
+            if not snap_text_columns_to_ink(fitz_page, page):
+                continue
+            for det, snapped in zip(columns, page.detections):
```
grossir: use zip(string=True)  to prevent silent failures (recycling) if the shapes are not as expected

scanning/services.py
```diff
@@ -1132,214 +1003,115 @@ def _compute_and_save_margin_rects(
...
+    :param llm_dir: The ``llm/`` directory of per-page PDFs.
+    :return: Number of files given a text layer.
+    """
+    if not settings.LLM_PAGE_TEXT_LAYER or not llm_dir.is_dir():
+        return 0
+    from blackletter.api import add_text_layer
+
+    _update_progress(scan_pk, "Adding a text layer to the LLM pages...")
+    added = add_text_layer(llm_dir)
+    print(f"  Text layer added to {len(added)} LLM pages", flush=True)
```
grossir: use a `logger.info`

scanning/services.py
```diff
@@ -1788,41 +1476,28 @@ def run_full_pipeline(scan_pk: int) -> None:
...
-
-        # 4. YOLO detection (all 3 models on bitonal)
+        # 3. YOLO detection (all 3 models on bitonal)
         _run_yolo(scan_pk, str(bitonal_path), str(output_dir))
 
-        # 5. Import detections into DB
+        # 4. Import detections into DB
         _update_progress(scan_pk, "Importing detections...")
         dets = _import_detections_from_json(scan_pk, str(output_dir))
+        _snap_text_columns_to_ink(scan_pk, str(bitonal_path))
```
grossir: Can this move off the upload path? Review 1 asks only whether the volume is complete and shows no detection overlay, so the column correction is review-2 prep — and it renders every page at 100 dpi to do it (~15 s on a 1,292-page volume). Suggestion is  dropping it here and doing it in run_generate_files alongside the rects, where the geometry is actually read.

scanning/services.py
```diff
@@ -1788,41 +1476,28 @@ def run_full_pipeline(scan_pk: int) -> None:
...
+        # 6. Pair opinions
         _update_progress(scan_pk, "Pairing opinions...")
         opinions = _re_pair_opinions(scan_pk)
 
-        # 8. Compute redaction rects (margin rects computed lazily when viewer requests them)
+        # 7. Compute redaction rects (margin rects computed lazily when viewer requests them)
         _update_progress(scan_pk, "Computing redaction rects...")
-        pdf_path = str(ocr_pdf) if ocr_pdf else scan.pdf_path
-        _compute_and_save_redaction_rects(scan_pk, pdf_path)
+        _compute_and_save_redaction_rects(scan_pk, str(bitonal_path))
```
grossir: Same question, bigger: this is ~2 more full-volume render passes before the scan flips to PENDING_REVIEW. Nothing in  review 1 reads redaction_rects — the overlay is step 2 — so a scanner waits ~45 s in total for geometry they won't see,  which cuts against what the Tesseract removal is buying.

I think these two steps belong in run_generate_files, next to the margins call you already added there. That keeps the  deliverable identical (Generate becomes self-sufficient) and takes the upload path down to the one pass  blackletter.api.pair does internally.

grossir (commented):
@quevon24 I think this is mostly ready to merge. This is a good step towards cleaning up Tesseract from the portal

I have released the paired blackletter version. So, to merge
- update the lock
- resolve merge conflicts
- a couple review comments relating to making the first processing group ligther, deferring redaction stuff until the first human review is completed (related to this https://github.com/freelawproject/scanning/issues/154)

This is a follow up issue https://github.com/freelawproject/scanning/issues/155. Text embedding was the major user of the Tesseract output. There are a couple smaller users we can get rid of, and the delete the Tesseract dependency completely

scanning/services.py
quevon24: Done in a52fad2.

scanning/services.py
quevon24: Done in a52fad2.

grossir (approved):
LGTM

PR #153, opened 2026-08-04
fix: make Recheck and the original-PDF views work without a local PDF

quevon24:
Fixes SCANNING-1S (https://freelawproject.sentry.io/issues/SCANNING-1S)

Clicking **Recheck** in step 1 raised `FileNotFoundError` from `Scan.pdf_path`. Production web pods have no local copy of the original PDF: S3 is the source of truth, and `PROCESSING_TMP_DIR` is only populated by the daemon or a lazy targeted pull.

## The recheck path never needed the PDF

`recalculate_issues` called `_parse_expected_range(scan.pdf_path)`, but that function only parses the *filename*, so requiring a local file bought nothing. Worse, the parse never worked for portal uploads: `book_upload_path` names files `<reporter>.<volume>.<start>.<end>.original.pdf`, and blackletter's `_infer_from_filename` only strips `.redacted`/`.ocr`/`.compressed`, so `int("original")` fails and it returns `(None, None)`. Verified against every naming path (`book_upload_path`, `_original_pdf_name`, pending-upload recovery, `make_dev_data`) plus Django's collision suffix. None of them parse.

Meanwhile the validate stage (`_dispatch_analyze`, `_rebuild_issues_from_results`) uses `scan.start_page` / `scan.end_page`. So Recheck was silently running with *no* expected range, discarding the missing-page, out-of-range, and mislabeled-document findings the original validation produced. `rebuild_page_map` (manual page-number edits) had the same problem behind a `try/except FileNotFoundError` whose fallback also parsed to `(None, None)`.

New `_expected_range(scan)` helper reads the DB fields, returning both values or neither. Two guards that checked only `exp_start` now require `exp_end` as well, since a scan with a `start_page` but no `end_page` would otherwise hit `None + 5` and `range(1, None)`. The `fitz.open(scan.pdf_path)` page-count refresh is now conditional on the file being present; the PDF has not changed during a recheck, and `page_count` is already stored.

**Behavior change:** rechecks now produce more issues than before on existing scans, because the range checks actually run. That matches the pipeline, so a recheck should now agree with the original validation instead of quietly dropping findings.

## Two consequences of turning the range back on

Giving `recalculate_issues` a real range woke up a ~45-line auto-correction block that had been dead code (with no range, `_split_detected` only ever flagged numbers below 1). It rewrites `ocr_results` and does not know about human input, so a curator who set PDF page 9 to `500` in the step-1 editor would click Recheck and see it silently changed to `9`, with a warning claiming "OCR read '500'". Manual reads (`zone`/`ocr` == `"manual"`, stamped by `assign_page`) are now skipped by the heuristic. They are still reported as out-of-range readings, so the curator gets a warning instead of losing their edit.

Recheck also did `scan.issues.all().delete()`, which took out `suppress_detection` rows. Those are curator decisions, not page-number findings, and the daemon rebuild already excludes them. `recalculate_issues` now does too.

## Same crash class in three other views

`export_pdf` was a guaranteed 500 in production for the same reason: `fitz.open(scan.pdf_path)` with no fallback, and `_is_preview_pdf` deliberately excludes the (up to 3 GB) original from what gets synced to a pod. `pair_opinions_api` and `compute_redactions_api` both evaluated `scan.pdf_path` *outside* their `try`, so a missing `bitonal.pdf` or OCR PDF meant an unhandled exception rather than the JSON error the viewer expects.

`serve_original_crop` already had the right pattern inline (catch, pull just the original, retry, 404). That is now extracted as `utils.local_original_pdf(scan)` and used by all four call sites. The pull stays targeted: `download_original_pdf` fetches only `*.original.pdf`, never the `images/` tree, into `PROCESSING_TMP_DIR`, and remains idempotent, so pods keep the same low-disk profile. Audited the rest: no unguarded `Scan.pdf_path` remains on any request path.

## Testing

`TestRecalculateIssues` and `TestRecalculateView` cover the recheck path (no local PDF, range from the scan, stray-reading flagging, no-`end_page` fallback, manual-edit preservation, suppression preservation). `TestLocalOriginalPdf` covers the helper's four branches, and `TestViewsWithoutLocalOriginal` covers the three endpoints. 351 tests pass.

Verified as real regression tests: 6 of the 8 original recheck tests fail against the parent commit, and the auto-correct pair pins the manual-edit guard from both sides.

## Follow-up, not in this PR

Blackletter's `mislabeled_document` message says "Filename suggests pages X to Y", but the range comes from the DB fields (and did in the pipeline all along). Fix belongs in the blackletter repo.

### DIFF
--- a/scanning/services.py
+++ b/scanning/services.py
@@ -1531,6 +1531,19 @@ def _expected_range(scan: "Scan") -> tuple[int | None, int | None]:
     return scan.start_page or 1, scan.end_page
 
 
+def _is_manual_read(result: dict) -> bool:
+    """Return whether a per-page page number was entered by hand.
+
+    ``assign_page`` stamps ``zone`` and ``ocr`` with ``"manual"`` when a
+    curator types a page number in step 1.
+
+    :param result: One per-page entry from ``Scan.ocr_results``.
+    :returns: ``True`` when the number came from a person, not a model.
+    :rtype: bool
+    """
+    return "manual" in (result.get("ocr"), result.get("zone"))
+
+
 def _split_detected(
     ocr_results: list, exp_start: int | None, exp_end: int | None
 ) -> tuple[list, dict]:
@@ -1675,6 +1688,11 @@ def recalculate_issues(scan: "Scan") -> None:
         in_range_sorted = sorted(in_range_by_page.items())
         offsets = {}
         for r in out_of_range:
+            if _is_manual_read(r):
+                # A curator typed this number, so it outranks the
+                # offset heuristic. It is still reported as an
+                # out-of-range reading below, just not overwritten.
+                continue
             p, detected = r["pdf_page"], int(r["detected"])
             before = [(pp, n) for pp, n in in_range_sorted if pp < p]
             after = [(pp, n) for pp, n in in_range_sorted if pp > p]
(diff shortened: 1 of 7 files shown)


grossir (approved):
LGTM

Issue #156, opened 2026-08-05
Daemon: batch-submit external jobs and poll them, instead of blocking on one scan at a time


quevon24:
## Current behaviour

`run_daemon` ticks `process_next_scan` every `DAEMON_POLL_INTERVAL` (default 5 s). That command recovers stale `PROCESSING` rows, then atomically claims **one** `QUEUED` scan with `SELECT ... FOR UPDATE SKIP LOCKED` and dispatches it.

Dispatch is synchronous all the way down: `services.py` calls `runpod_client.detect()` / `.analyze()`, and `runpod_client._invoke` submits to `/run` and then **blocks polling `/status`** until the job reaches a terminal state or `RUNPOD_REQUEST_TIMEOUT` (default 1800 s) expires.

So a daemon process holds exactly one scan in flight and spends most of its time blocked on a remote GPU job. Ten queued volumes are processed strictly one after another.

## Why change it

Two separate problems, worth keeping apart because they have different fixes.

**Latency.** Ten volumes take ten times as long as one, even though the GPU provider would happily run them concurrently. Scanners should be able to get feedback the same day they upload.

**Cost.** Each job pays its own cold start and its own idle tail. Measured on a live endpoint (see #147): cold start ~124 s, and on the 24 GB tier a 300 s idle timeout costs $0.058 against $0.031 of actual work for a small job — the tail can cost more than the work it follows. Modelling in #147 put combined overhead at roughly 12% of the bill under the current serial flow, dropping to about 4% once jobs are batched against a deliberately narrow worker pool.

**Batching alone does not save money.** Submitting a wide burst against an uncapped pool costs the same as today, because every job still lands on its own cold worker. The saving comes from capping the endpoint's max worker count so each warm worker absorbs several jobs back-to-back — queue time is free, so a narrow pool is strictly cheaper at the cost of wall clock. Batching buys latency; the cap buys cost. Both are wanted, but they are independent, and the cap is an endpoint console setting rather than code.

## Proposed shape

Replace "claim one scan, block on it" with a **batch cycle**:

1. **Claim a batch.** Snapshot the scans that are queued right now. Scans queued *after* the snapshot are deliberately ignored and picked up by the next cycle — a closed batch makes "this cycle is finished" a well-defined condition and stops a busy queue from extending a cycle indefinitely.
[...]

Issue #150, opened 2026-08-03
Generalize the external job layer for per-scan batches and hosted API providers


grossir:
We will use 
- serverless or dedicated Runpod to run dots.mocr. A scan will be sharded for speed
- serverless or dedicated Runpod to run  YOLO model (to decide [here](https://github.com/freelawproject/scanning/issues/147))
- the Mistral OCR API

We need a separate model to track these jobs. A lot of these fields are inside the Scan model because we expected a sequential runpod jobs per scan. The model should support:
- tracking the rundpod job id for resuming after the daemon has been interrupted. See https://github.com/freelawproject/scanning/issues/127
- we need to experiment with the Mistral API to see if it needs any special handling. Should this be a separate model sharing an abstract model with the Runpod jobs?
- a status field so the daemon can poll it / resume, etc. 
- How will the above interact with the Scan status? We will also need to redirect the requeue logic from the admin
- a field to hold the S3 response reference
- a way to identify how the shard was constructed. This will be useful for rebuilding the whole PDF in the future. Note that the way we do this must handle the case when the Scan may have a page deleted or added, so it can't be raw page indexes

About deleting the duplicated the fields that will be "duplicated" like retry_count or interruption_count from the Scan model, I think that should wait for another cleanup issue and PR

A note on sharding the Scan for dots.mocr (and other OCRs): 
- The main reason is processing parelllism speed. We want the scanners to be able to get feed the same day they upload the original scan
- It has other benefits, like error handling / retrying
- Is there a response size limit? If it exists, we may need to compute the size of the shard itself.
- Right now the RunPod image gets the PDF from S3. Since it will be splitted, can we avoid each partial job downloading the whole file? Is there a safe way? Worth to research this. If the sharding happens on the daemon, how long will it take?  This may be a separate sub issue

Related work:
- frontend changes to handle these model changes (may be a separate PR as long as we don't break the portal on merging)
- once the model is in, do we need a data migration for the fields that were contained in the Scan?
- a migration to clean the Scan model out of these job fields

quevon24:
## Proposal: one `ExternalJob` table, plus a worker contract change that has to land first

Design pass over what this issue asks for, given how the daemon and the workers look today. The model draft is in section 4; the sections before it are the constraints that shaped it. The draft passes `manage.py check` and `makemigrations` against a stubbed app, so field defs, choices, and index names are at least mechanically sound.

**Updated** after further design discussion, in case anyone read the first version:
[...]

PR #159, opened 2026-08-06
feat(scanning): return worker outputs via presigned PUT to S3

### LINKED ISSUE #157: Return worker outputs via presigned PUT to S3 instead of inline in the job response
Sub-issue of #147.

## Why

Workers return their whole payload inline in RunPod's job output today. Two consequences:

- The ~20 MB response cap becomes a design constraint (item 4 of #147). It is the reason we cannot return per-word bounding boxes, which is the data the extraction pipeline exists to produce.
- The payload only lives as long as RunPod's async result retention (~30 min). After that `/status` returns 404 and `runpod_client._poll` (see the 404 branch, `scanning/runpod_client.py:549`) throws the finished work away and resubmits the whole job.

Proposal: the daemon computes the result key, presigns a **PUT**, and passes it in the job input alongside the existing presigned GET. The worker uploads results straight to S3 and the response carries only the key plus timings and per-unit success/failure. Job status stays authoritative for whether a job finished. No browser is involved, so CORS is not a concern.

One correction to the premise in [grossir's comment](https://github.com/freelawproject/scanning/issues/147#issuecomment-5184792964): inline results are *less* durable than they look. They are gone after RunPod's retention window, and we already discard them on a 404. An S3 object outlives both the retention window and the daemon process, so this change makes recovery strictly better, not worse. That is most of the answer to both questions below.

## Contract change

- Job input gains `result_url` (presigned PUT, one object) and `result_key`.
- Worker writes one JSON object with a single PUT, then returns `{"result_key": ..., "bytes": n, "sha256": ..., "duration_ms": ..., "failed": [...]}`.
- If `result_url` is absent the worker returns inline exactly as today. That keeps local/dev mode and the existing tests working, and lets the worker side land before the daemon side.
- New structured error codes so the transient/terminal classification in `_TRANSIENT_ERROR_CODES` stays explicit rather than falling through to the generic FAILED branch: `RESULT_UPLOAD_FAILED` (transient) and `RESULT_URL_EXPIRED` (transient, but must not reuse the URL).

## What happens if the upload fails?

- The worker retries the PUT with backoff before giving up. The compute is already paid for, so a network blip should not cost a re-run.
- HTTP 403 means the signature expired. Retrying the same URL cannot succeed, so the worker fails fast with `RESULT_URL_EXPIRED`, mirroring how `_download_pdf` already treats a 403 on the input URL.
- A job that fails never writes its object. The job goes FAILED, existing transient handling resubmits with a fresh presign, and because result keys are run-scoped nothing stale is left behind for the retry to trip over.
- Deliberately no inline fallback. Carrying the payload in the response "just in case" reinstates the size cap and creates two sources of truth for the same result.
- Once #150 shards a scan, the unit of loss is one shard, not the volume.

## How do we find results if the watcher dies, and how do we know they are complete?

We should not sniff or list an S3 prefix. A presigned PUT covers exactly one object and a prefix cannot be presigned, so the key has to be known at submit time regardless. It is deterministic, computed daemon-side, and persisted on the job row before submission (#150), which turns discovery into a `head_object` on a known key instead of a search.

Recovery order, with job status still the authority:

1. `/status` says COMPLETED: read `result_key`.
2. `/status` returns 404, or the daemon died mid-flight and is resuming from job rows: `head_object(result_key)`. Present means harvest the finished work. Absent means the job did not get far enough, so resubmit with a fresh presigned URL.
3. `/status` says FAILED and the object is absent: the existing retry path, unchanged.

Polling S3 alone is never sufficient, because a failed job also writes nothing, so absence cannot distinguish "still queued" from "failed" and every failure would hang until timeout.

Completeness rests on three properties rather than a heuristic:

- **A single PUT is atomic.** A partial upload never becomes a gettable object, so "the object exists" implies "all of its bytes are there". Keeping results to one PUT (no multipart) keeps that simple; shard sizing should stay well under the 5 GB single-PUT limit, which is not a real constraint for JSON.
- **Keys are run-scoped and never overwritten** (`processing/{pk}/.../jobs/{stage}/{provider}/run{run}/shard_{i:04d}.json`). Without that, a `head_object` probe could harvest a previous run's output and report it as this run's result.
- **The payload is self-describing.** Schema version, shard index, source digest, and expected unit count, validated on consume. A truncated or wrong-generation file then fails validation loudly instead of being consumed as a legitimately short result.

The presigned PUT's lifetime has to outlive queue time plus execution. `RUNPOD_PRESIGNED_TTL` is 86400 s today, generous for a single job, and batch submission (#156) makes that margin load-bearing rather than precautionary since a burst can sit queued far longer. The job deadline should sit well inside the URL's TTL, so we cancel before the signature expires rather than paying for compute that cannot be delivered.

## Also in scope

- **Cleanup.** Run-scoped keys accumulate orphans from re-runs and from jobs whose results were never consumed. The consumed data lives in Postgres, so an expiry lifecycle rule on the job-results prefix is enough, plus prefix deletion when a scan is deleted.
- **Security posture.** A presigned PUT is a write capability handed to a third-party GPU host. It stays scoped to one key, one method, and a bounded TTL, and it preserves the property our existing worker image deliberately has: no AWS credentials baked into the image. Scoped temporary credentials would be more flexible but would give that up.
- **Tests.** A stubbed PUT target covering success, 403, and 5xx-with-retry; the inline path when `result_url` is absent; and the `head_object` recovery branch on a `/status` 404.

## Out of scope

- The job row that stores the key, shard identity, deadline, and retry counters: #150.
- Batch submission and the non-blocking sweep: #156.
- Pod vs serverless, engine sizing, and the render-format flag: #147.
- Published extraction file naming: #150.

quevon24:
Fixes #157.

First implementation of presigned PUT result delivery. Deliberately minimal: it changes how a job's payload gets back to the daemon and nothing else, so the pipeline, the daemon loop and the services layer all behave exactly as they do today. Most of the logic here should be reusable as-is when the job row (#150) and batch submission (#156) land.

## What changes

The daemon presigns a PUT alongside the presigned GET it already sends, and passes `result_url` + `result_key` in the job input. The worker writes one JSON object with a single PUT and returns only status, key, size, digest and timings. The daemon polls `/status` exactly as before and, on COMPLETED, pulls the object and validates it instead of reading the payload out of the response.

Why: RunPod caps a job response at ~20 MB (the reason we cannot return per-word bounding boxes) and discards it once the ~30 min retention window closes. An S3 object has neither limit.

`detect()` and `analyze()` return the same shape they always have, so `services.py` is untouched.

## Result keys and staleness

Keys are per scan and action (`processing/{pk}/.../jobs/{action}/result.json`), not per run, so a re-run overwrites its predecessor rather than accumulating orphans.

Because the key is reused, nothing is read without checking the object was written after the reading job was submitted (`head_object` on `LastModified`, 60s clock skew allowance), and the envelope's `job_id` is checked against the job we polled. The time check is the cheap first pass; the job id is authoritative.

A 404 from `/status` now probes that key before giving up, so a job that finished and uploaded before RunPod dropped its record is harvested instead of being paid for twice. Any S3 error during that probe is swallowed and the scan re-queues as before.

## Backwards compatibility

Nothing happens without AWS credentials, so local dev and CI keep using the in-process blackletter path untouched.

If `result_url` is absent the worker answers inline exactly as today, and the daemon accepts an inline response even when it sent a `result_url`. That means rolling the worker image back is a complete rollback, with no daemon change.

## S3 region (needs a decision before merge)

Nothing in this project set an S3 region, so boto3 fell back to `us-east-1`. That was invisible until now because SigV2 presigned URLs carry no region and S3 redirects a misdirected GET. These URLs are SigV4 (needed so the signature covers the PUT's `Content-Type`), and SigV4 signs the region into the credential scope, so a mismatch fails outright with `AuthorizationQueryParametersError`.

Added `AWS_S3_REGION_NAME`, defaulting to **`us-west-2`**, which is where the dev buckets live (confirmed by hitting the error against `dev-com-freelawproject-scanning-private-storage`).

Two things to check before this merges:

- Confirm the production buckets are also `us-west-2`: `aws s3api get-bucket-location --bucket <name>`. Override with the env var if not.
- That setting name is the one django-storages reads, so it now pins the region for ordinary media uploads too. Intentional (one region for every S3 client beats two ways to be wrong) and it should be strictly more correct, but it does touch the upload path. Happy to scope it to the RunPod client only if reviewers prefer.

## Error codes

| `error_code` | Cause | Daemon behaviour |
|---|---|---|
| `RESULT_UPLOAD_FAILED` | PUT failed after retries (5xx, 429, dropped connections) | Re-queue (transient) |
| `RESULT_URL_EXPIRED` | Presigned PUT dead (HTTP 403) | Re-queue; resubmitting mints a fresh signature |
| `RESULT_UPLOAD_REJECTED` | S3 refused the write for a reason a re-run cannot fix (wrong signing region, bucket policy) | Mark scan ERROR |

The last one is terminal on purpose: those are configuration errors, and treating them as transient re-runs the GPU work, and re-bills it, once per retry before failing with the same message.

## Testing

Full suite green (406 tests, +40 here). Covers the upload retry ladder, 403 and 4xx handling, the inline path, envelope validation, the freshness and job-id checks, and the 404 probe.

Verified end to end against a real endpoint (image on a personal Docker Hub repo, RunPod serverless, dev bucket):

- Job response carried no `detections`, only `{"status": "succeed", "result_key", "bytes": 4956, "sha256", "upload_ms": 695, ...}`.
- Object present at the expected key, `ContentType: application/json`, `ContentLength` matching the reported `bytes`.
- `sha256sum` of the downloaded object identical to the digest the worker reported.
- Envelope `job_id` identical to the RunPod job id, 23 detections across 15 pages in the payload.

## Out of scope

Persisting the job id and result key on a job row so a restarted daemon can resume (#150), the non-blocking sweep and batch submission (#156), sharding and per-shard partial failure, and the S3 lifecycle expiry rule on the results prefix (infra config, not yet applied).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -15,6 +15,10 @@ DB_SSL_MODE=prefer
 # AWS_ACCESS_KEY_ID=
 # AWS_SECRET_ACCESS_KEY=
 # AWS_STORAGE_BUCKET_NAME=
+# Region both buckets live in. Presigned URLs are SigV4, which signs
+# the region into the credential scope, so this must match or S3
+# rejects them. Default us-west-2.
+# AWS_S3_REGION_NAME=us-west-2
 
 # Sentry
 # SENTRY_DSN=
@@ -37,9 +41,10 @@ DB_SSL_MODE=prefer
 # Transport-error retries on /run (network blips, 5xx). Terminal job
 # failures are never retried. Default 2.
 # RUNPOD_MAX_RETRIES=2
-# Lifetime of presigned GET URLs handed to the worker (seconds).
-# Default 86400 (1 day); generous headroom so a URL never expires
-# before the worker fetches it. AWS SigV4 max is 7 days.
+# Lifetime of the presigned URLs handed to the worker (seconds), both
+# the GET for the input PDF and the PUT for the result. Default 86400
+# (1 day); generous headroom so neither expires before the worker gets
+# to it. AWS SigV4 max is 7 days.
 # RUNPOD_PRESIGNED_TTL=86400
 # Max transient RunPod failures (NO_GPU, 404) before a scan escalates
 # to ERROR instead of being re-queued. Default 5.
--- a/scanning/runpod/README.md
+++ b/scanning/runpod/README.md
@@ -21,11 +21,18 @@ in RunPod endpoint env vars, never in the image.
 │                     │  POST  │                      │
 │ runpod_client.py    ├───────▶│ handler.py           │
 │  - upload PDF to S3 │  /run  │  - download PDF via  │
-│  - presign GET URL  │        │    presigned URL     │
+│  - presign GET + PUT│        │    presigned GET     │
 │  - submit job       │        │  - blackletter.api   │
 │  - poll /status     │◀───────│    .detect / analyze │
-│  - return result    │        │  - return JSON       │
-└─────────────────────┘  JSON  └──────────────────────┘
+│  - read result key  │  key   │  - PUT result to S3  │
+│  - fetch from S3    │        │  - return the key    │
+└─────────────────────┘        └──────────────────────┘
+          │                               │
+          │        ┌──────────────┐       │
+          └───────▶│      S3      │◀──────┘
+             GET   │  PDF in,     │  PUT
+                   │  result out  │
+                   └──────────────┘
 ```
 
 Full design rationale, payload-size math, and tradeoffs are in
@@ -433,8 +440,10 @@ pod, the lowest-friction option is #3 — no code changes required.
 ## Result retention
 
 RunPod purges job records after a short window. You **must** poll
-`/status` and persist the output before this window expires, or the
-result is lost.
+`/status` before it expires to learn a job's fate. The *payload*
+survives longer than that when the worker writes it to S3 (see
+[Result delivery](#result-delivery)); only the job record itself is
+subject to the retention below.
 
 | Submission path | Result retention |
 |---|---|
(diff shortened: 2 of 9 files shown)


grossir (approved):
LGTM!

PR #161, opened 2026-08-07
Resume the pipeline on a re-queue instead of re-running the GPU stages

### LINKED ISSUE #160: Skip completed pipeline steps on a re-queue instead of re-running YOLO detection
## What happened

Scan 1946, 2026-08-07. The daemon was restarted mid-pipeline for a deploy. Detection had already completed, been harvested, and been imported into the DB:

```
11:24:32  runpod detect job 75a75765... COMPLETED in 194297ms (action=186197ms, large=66032ms, medium=65582ms, small=54153ms)
11:24:32  harvested detect result from s3 key=<processing_prefix>jobs/detect/result.json (2625251 bytes)
11:24:32  YOLO detection done (213.7s)
11:24:36  PaddleOCR validation...
11:26:14  Received signal 15, shutting down...
11:26:14  Re-queued 1 interrupted scan(s) for the next daemon tick.
11:27:43  Daemon started. Schedule: process_next_scan=5s, cleanup_processing_tmp=900s
11:27:43  Processing scan 1946 (full_pipeline)
11:27:51  Downloaded 3 processing file(s) for scan 1946
11:27:52  YOLO detection...
11:27:52  runpod detect job 03a3b80e... submitted
```

The second job recomputed identical detections and overwrote the `result.json` that already held them, at the cost of another ~194s of GPU time.

## Why

`run_full_pipeline` has no resume point. A re-queue restarts it at step 1.

Step 2 looks resumable only because `_ensure_bitonal` guards on file existence (`scanning/services.py:397-398`), so `bitonal.pdf` was skipped. Step 3 has no such guard: `_run_yolo` is called unconditionally (`scanning/services.py:1545`) and writes `detections.json` only after the job returns (`scanning/services.py:498`).

Three copies of the finished work were available at restart and none was consulted:

- `Detection` rows in the DB. The first run reached step 5, past step 4's `_import_detections_from_json`, and `run_paddleocr_validation` does not delete them.
- `<processing_prefix>jobs/detect/result.json` in S3, 2.6 MB, harvested at 11:24:32.
- `detections.json` on local disk, which does not survive a fresh pod. `_run_yolo` writes it locally and never uploads it; only `_sync_detections_to_disk` pushes it (`scanning/services.py:809`), later in the pipeline. The restart's `Downloaded 3 processing file(s)` line confirms it was not in S3 to pull back.

The analyze step has the same gap. Job `66880db1` went IN_PROGRESS at 11:24:36 and was abandoned at 11:26:14. It most likely completed on RunPod and PUT its result to `<processing_prefix>jobs/analyze/result.json` with nobody left to harvest it.

This predates #146; that PR only removed the OCR pre-pass, whose `find_ocr_pdf` check happened to make it resumable. Detection never was.

## What should change

The rule we want: a step re-runs and overwrites its S3 output only when we intended to re-run it. An interrupted pipeline picking up where it left off is not an intent to re-run.

No model change is needed. The freshness check already exists in the shape we want: `runpod_client._result_object_is_fresh` (`scanning/runpod_client.py:503`) does one `head_object` and compares `LastModified` against a reference time.

- In `run_full_pipeline` step 3, skip the detect submit when `head(<processing_prefix>jobs/detect/result.json).LastModified` is newer than `head(<processing_prefix>bitonal.pdf).LastModified`, and rehydrate `detections.json` from the stored result instead of re-running the job.
- Use `bitonal.pdf`'s S3 timestamp as the reference, not a `Scan` field. `date_modified` is `auto_now=True` (`scanning/models.py:157`) and `AutoNowQuerySet.update()` stamps it on every bulk write (`scanning/models.py:33-44`), so `_update_progress` bumps it on every progress tick. `processed_at` is only set at the end of a run, so it is null exactly when a resume needs it.
- Stop `upload_processing_files` from re-uploading an unchanged `bitonal.pdf`. It currently overwrites with no hash check (`scanning/s3_sync.py:229-230, 258`) and `_push_processing_files_to_s3` runs at the end of five pipelines (`scanning/services.py:1495, 1571, 1759, 2150, 2474`), so the reference timestamp would advance even when the bytes did not, and the guard would never fire. Skipping when the local size matches the S3 head is enough; `_ensure_bitonal` already pushes the file on creation (`scanning/services.py:413`).
- Add an explicit force/rerun flag defaulting to off on the resume path. Keep `run_detect` (`scanning/services.py:2442`) and the other explicit entry points forcing a real re-run, since re-detect is their whole purpose. The guard belongs at the `run_full_pipeline` call site, not inside `_run_yolo`.
- Give the analyze step in step 5 the same treatment.

The result objects themselves need no protection: `upload_processing_files` already skips the whole `jobs/` subtree (`_is_job_result`, `scanning/s3_sync.py:165, 255-256`).

## Scope

This is deliberately a stopgap until #150 lands the external job layer, where a job row records which stage completed against which input and the sweep daemon harvests in-flight jobs instead of orphaning them.

It should not grow into that design. What it wants is the smallest guard that stops paying for duplicate GPU jobs and stops clobbering good S3 results on every daemon restart.

quevon24:
Fixes #160.

## Why

The daemon takes a SIGTERM on every deploy and re-queues whatever was mid-flight, but `run_full_pipeline` restarts at step 1. Step 2 looked resumable only because `_ensure_bitonal` guards on file existence; step 3 had no such guard, so a detect that had already completed was re-submitted. On scan 1946 that was another ~194s of GPU time to recompute detections that were already sitting in S3, and the result object was overwritten with an identical one.

The finished work was available in three places at restart and none was consulted. The DB still had the `Detection` rows, S3 still had the 2.6 MB result object, and `detections.json` was gone because `_run_yolo` writes it locally and never uploads it.

## What changed

**Reuse is a timestamp comparison, not new state.** `runpod_client.reusable_result` heads the stored result and the stage's input PDF, and hands back the payload only when the result is newer. No migration, no job table. It deliberately ignores the envelope's `job_id`, which is the inverse of `_validate_envelope`: there the question is "did *this* run write it", here we are knowingly reading an earlier run's output.

**The uploader had to stop touching the reference timestamp.** `upload_processing_files` overwrote with no hash check and runs at the end of five pipelines, so it advanced `bitonal.pdf`'s `LastModified` on every run even when the bytes were identical, and the guard would never have fired. It now skips a stage input whose remote size already matches.

That skip covers `*.original.pdf` as well as `bitonal.pdf`, which is broader than the issue's bullet named. Step 5 analyzes the original rather than the bitonal, and the original also lives in the output dir, so the analyze half needed the same protection.

**Re-validate deletes the stored results.** This is the part the issue got wrong and a review caught. The issue asked for a `force` flag, and a flag is unusable here: the daemon dispatches by name with only a pk, so nothing can set it. Left as designed, the "Re-validate" button (labelled "re-run the full pipeline from scratch", warning about GPU cost, behind a confirm dialog) would have silently handed back the detections the scan already had, and `run_detect` was no escape hatch because `start_detect` short-circuits to step 2 whenever detections exist. `s3_sync.invalidate_job_results` expresses the intent where the user expresses it, and survives the hand-off through the queue in a way a parameter cannot.

Deliberately not wired into the admin re-queue: that action documents itself as recovery for a scan stuck behind a killed daemon, which is exactly the resume case this PR exists to make cheap.

## Scope

A stopgap ahead of #150, which replaces this with a job row recording which stage completed against which input, plus a sweep daemon that harvests in-flight jobs instead of orphaning them. Job `66880db1` in the same incident went IN_PROGRESS at 11:24:36 and was abandoned at 11:26:14; it very likely completed on RunPod and PUT its result with nobody left to read it. That orphaning is not fixed here.

Two limits worth knowing. The freshness check is keyed only on the input's timestamp, so it is closer to "has this stage ever completed for this scan" than to a resume marker, which means new model weights will not propagate to a scan that already has a result until someone hits Re-validate. And the upload skip compares size rather than a hash, to avoid digesting a 60+ MB PDF at the end of every pipeline; a same-size edit would go unnoticed, so `_reuse_detect_result` also rejects any result referencing a page past the current page count.

## Also here

`requests` is now declared in base deps. `runpod_client` imports it directly, but it was only ever arriving as a transitive dependency of the `local-ml` extra, so a plain `uv sync` (what the lean web image builds from) resolved no `requests` at all. Nothing was broken in practice, since every import of `runpod_client` is function-local and all five sit on daemon paths, but the declaration was wrong. Verified with `uv export --no-dev --no-emit-project`, which now yields `requests==2.33.1` with no `ultralytics` or `paddleocr` present.

## Testing

488 tests, up from 466. New coverage: the freshness comparison including the older-than-input rejection and the foreign-`job_id` acceptance, the exact S3 keys both stages head, the upload skip and its re-upload case, the page-count backstop, and that a blocked Re-validate does not throw away a reusable result.

The reuse path has not been exercised against real S3; everything here is unit-tested with boto3 mocked, so the first production re-queue is the real test.


grossir (approved):
LGTM

Issue #158, opened 2026-08-06
Evaluate using doctor for bitonal


grossir:
Sub-issue of #152.

#152 concluded bitonal stays (best size at roughly equal speed) and does not need to be the model input. Open question: run it in doctor instead of on the scanning pod.

### Why offload

Bitonal is ~6.8 min of CPU per volume on the portal host (1293 pages @ 315 ms/page, #87), competing with the web process. blackletter already ships a forked converter (2.9× at 4 workers), but fanning out on the pod starves reviewers. Doctor is a separate, horizontally-scaled host — the lowest-friction way to move the CPU off the pod. Nothing else has to move with it: detection boxes are page-rect-relative, so bitonal can stop being the model input without touching geometry (#152).

Scope note: bitonal is the first of ~4 full-volume raster passes on the pod. The others (margins `ink.page_mask`, redaction `grow_to_ink`, generate/split) need blackletter geometry plus detections and are a separate question — doctor is not their destination.

### Reuse the dots.mocr shards

Reuse the shard *definition* from #150 — one definition, N jobs: `(scan, engine, shard)` with `engine=bitonal, provider=doctor` alongside `engine=dots, provider=runpod`. A 200-page shard is ~45 s on production CPU, short enough not to camp a gunicorn worker.

Pre-split the volume into per-shard PDFs in S3 once (stream copy, no rasterization) so no job downloads the whole volume — answers #150's open question for both engines at once.

Invariants, because bitonal is the one 1:1 page-preserving consumer:

- shard set covers every page, no gaps, no reordering (both engines run pre-review-1 on the full page set; if dots moves to the clean set this reuse breaks)
- merge in strict shard order
- assert `merged.page_count == original.page_count` and per-page rect equality

Shards die at merge: `bitonal.pdf` is then an ordinary PDF, so page inserts/deletes work as they do today and shard identity never has to survive a page edit. An inserted page becomes a single-page doctor call — an improvement, since today the raw image is stamped into `bitonal.pdf` unconverted (`services.py:1791`). Its page rect must come from the neighbour page, not from image dims.

### Presigned GET / PUT

- Input: presigned GET of the shard PDF.
- Output per #157: two PUTs — `shard_NNNN.pdf`, then a status object written **last**. Existence of the status object means complete; each PUT is atomic; keys are run/attempt scoped and never overwritten.
- Status payload records `dpi`, `threshold`, page range, source digest. The parameters are still in flux and a stale-DPI chunk must not be consumable.
[...]

PR #162, opened 2026-08-10
perf(scanning): run bitonal conversion across multiple workers

quevon24:
## Blocked until blackletter ships

This needs [blackletter#72](https://github.com/freelawproject/blackletter/pull/72) merged **and** a version tagged and published to PyPI. The daemon passes a `workers` argument to `api.bitonal` that no released blackletter accepts.

`pyproject.toml` is therefore pinned to a commit rather than a release, in both the base dependency and the `local-ml` extra:

```toml
"blackletter @ git+https://github.com/freelawproject/blackletter@a6e905fcb585b624a970fef2d178aeca89241f38",
```

That commit lives on the blackletter PR branch, so it stops being reachable if the branch is deleted on merge. **Re-pin to `blackletter>=0.3.0` (or whatever gets tagged) before this merges.** Do not land it with the git pin in place.

## What this does

Bitonal conversion ran sequentially in one process. blackletter can now split it across workers, and this side chooses how many, because only the deployment knows what the pod was granted.

`BITONAL_WORKERS` derives from the cores actually available, leaving about 20% headroom:

```python
BITONAL_WORKERS = env.int("BITONAL_WORKERS", default=0) or max(
    1, int(_available_cpus() * BITONAL_CPU_FRACTION)   # 0.8
)
```

`_available_cpus()` intersects `sched_getaffinity` with the cgroup quota (v2 `cpu.max`, falling back to the v1 pair). `os.cpu_count()` reports the *node's* cores inside a pod, which oversubscribes badly. Derivation checked at 22 cores to 17, **8 cores to 6**, 2 to 1, 1 to 1, and an explicit `BITONAL_WORKERS=3` to 3.

Under `TESTING` it is forced to 1, so a test that reaches a real conversion cannot spawn a pool sized to whatever CI happens to be running on.

## Measured on a real volume

Through the daemon, on a 929-page volume (1.33 GB source), before and after on identical input:

| bitonal conversion | time |
| --- | --- |
| before | 337.6s |
| after, 17 workers | 33.8s |

**10x**, same 68.3 MB output. On the 8-CPU pod expect roughly 65s rather than 33.8s.

Full pipeline for that scan is now 713s, of which bitonal is 4.7%. RunPod detection plus validation is 554s, or 77.7% of the run, so that is where any further pipeline work belongs rather than here.

## Entrypoint

`docker-entrypoint.sh` now logs `Installing local blackletter...` when it installs from the `/opt/blackletter` mount. Without it there is no way to tell from the logs whether a container is running the pinned package or a local checkout, which cost me a wrong diagnosis while testing this: a container silently running mounted code looks identical to one running the pinned wheel.

Worth knowing, since it interacts directly with the pin above: when that mount is present the editable install shadows whatever `pyproject.toml` pins. So in a dev container the pin is inert and the local checkout wins. That is the intended workflow for iterating on blackletter, but it means dev is not a test of the pin. Verified separately: a clean venv installing from the pinned URL, non-editable and with no mount, does expose `workers`.

The entrypoint only logs. It does not create the mount, which comes from `docker-compose.yml` and is not part of this PR, so the new line only fires for a developer who has added that volume themselves.

## Testing

Two new tests: one asserting the kwarg actually reaches blackletter (a dropped kwarg costs the entire speedup while failing nothing), one on the derivation bounds.

Full suite passes against the real pinned package rather than a mock: 461 passed, 1 skipped.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -20,6 +20,12 @@ DB_SSL_MODE=prefer
 # rejects them. Default us-west-2.
 # AWS_S3_REGION_NAME=us-west-2
 
+# Processing
+# Worker processes bitonal conversion is split across. Unset (the default)
+# derives it from the cores the pod was actually granted, leaving ~20% of
+# them free. Set it to pin the count, or to 1 to convert inline.
+# BITONAL_WORKERS=
+
 # Sentry
 # SENTRY_DSN=
 
--- a/docker/django/docker-entrypoint.sh
+++ b/docker/django/docker-entrypoint.sh
@@ -5,6 +5,7 @@ case "$1" in
 'web-dev')
     # Install blackletter from local mount if available
     if [ -d /opt/blackletter ]; then
+        echo "Installing local blackletter..."
         uv pip install -e /opt/blackletter 2>/dev/null || pip install -e /opt/blackletter 2>/dev/null || true
     fi
     python manage.py migrate
@@ -19,6 +20,7 @@ case "$1" in
 'run_daemon')
     # Install blackletter from local mount if available
     if [ -d /opt/blackletter ]; then
+        echo "Installing local blackletter..."
         uv pip install -e /opt/blackletter 2>/dev/null || pip install -e /opt/blackletter 2>/dev/null || true
     fi
     exec python manage.py run_daemon
--- a/docker/django/docker-entrypoint.sh
+++ b/docker/django/docker-entrypoint.sh
@@ -1,13 +1,20 @@
 #!/bin/sh
 set -e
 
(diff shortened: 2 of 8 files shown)


grossir:
Parallelization assumes the pod has 8 CPUs. The node does. The pod has no resource limits. We would be taking over the whole node (shared with courtlistener pods) just for bitonal. Related to https://github.com/freelawproject/blackletter/issues/71#issuecomment-5273821395

Apart from the parallelization speed will be gained from #158 without taking over the pod

Closing this

Issue #165, opened 2026-08-11
Preserve OpinionScan rows across a re-generate instead of deleting and recreating them


quevon24:
## What happens now

`run_generate_files` throws away every opinion row for the scan and rebuilds them from scratch:

```python
OpinionScan.objects.filter(scan=scan).delete()
for i, op in enumerate(existing_opinions):
    opinion = OpinionScan.objects.create(...)
```

(`scanning/services.py:2103`)

The recreate sets `scan`, `reporter`, `volume`, `opinion_order`, `page_start`, `page_end`, `caption_page_index`, `key_page_index`, `has_image`, `uploaded_by`, and hard-codes `status=OpinionStatus.OK`. Everything else about the row is gone.

Regenerating files is a routine action: fix one redaction box, or re-pair a single opinion, and the whole set is rebuilt.

## Why it matters

**1. It resets review state.** `status` goes back to `OK` on every regenerate, so an opinion a human marked `GAP` or `ERROR` silently reads as fine again, and `notes` and `process_output` are dropped. `status` and `notes` are editable in the admin and `status` is a filter on the opinion list, so a reviewer's decision is being discarded with no trace.

**2. It changes every primary key.** `opinion_detail` routes by pk, so any link or bookmark to an opinion breaks on each regenerate.

**3. Anything we later attach per opinion inherits the problem.** The per-opinion OCR planned in #150 sends each opinion PDF to several engines, so a 300-opinion volume implies hundreds of records and results hanging off these rows. Whatever shape those take, they cannot outlive a row that is deleted and recreated, and re-running that work costs real money. Worth fixing before anything expensive depends on opinion rows being stable.

## What to do

Replace delete-all-and-recreate with update-or-create, so a row survives when its opinion did not change.

The real work is choosing the identity to match on:

- **Page range (`page_start`, `page_end`) looks right.** It is the natural key of an opinion, it is what `opinion_pdf_path` already embeds in the filename, and it changes exactly when the opinion genuinely changed. An unchanged range keeps its row and its review state; a changed range is a different opinion, and replacing it is the correct outcome.
- **`opinion_order` is not usable** as the key, because inserting or dropping one opinion renumbers every later one.

Two things worth handling in the same change:

- Rows whose range no longer appears must still be deleted, so a merged or removed opinion does not linger.
[...]

PR #166, opened 2026-08-11
feat(scanning): add ExternalJob to track work sent to providers

### LINKED ISSUE #150: Generalize the external job layer for per-scan batches and hosted API providers

quevon24:
## Fixes

Part of #150. Deliberately does not close it: this is the model only. The integration, the admin re-queue redirect, the frontend, and dropping the now-duplicated `Scan` fields all stay with that issue.

## Summary

One new table, `ExternalJob`, plus its admin, factory, and tests. Nothing calls it yet, so there is no behaviour change and the migration is a single additive `CreateModel`. No data migration is needed either, because #136 never landed, so there are no `Scan.runpod_job_*` fields to copy across.

A job row is one unit of work handed to an external provider. The daemon keeps no in-memory job state: everything it needs to decide what to submit, poll, harvest, or cancel is on these rows, which is what makes an interrupted daemon resumable rather than a special case.

### The decisions worth pushing back on

**`provider` and `engine` are separate columns.** `provider` (runpod, mistral) decides the transport and how progress is discovered. `engine` (blackletter, dots_mocr, mistral_ocr, surya, lighton_ocr) decides what the work is. With blackletter alone this looks like dead weight, but it is what lets several models share one polling implementation, and `engine` is part of the unique key, so it is the one thing that would be awkward to add later.

**Stages come in two shapes, and the database enforces which.** `DETECT` and `ANALYZE` run once over the volume, before review. `EXTRACT` and `TIEBREAK` run after file generation, once per opinion PDF, so a 300-opinion volume read by three engines is 900 rows. Hence:

- a nullable `opinion` FK, with `scan` still non-null and denormalized, because "has this scan anything in flight" is the daemon's most frequent question and should not need a join
- two *conditional* unique constraints. Volume rows are unique on `(scan, stage, engine, run, shard_index)`, opinion rows on `(opinion, stage, engine, run, shard_index)`. They have to be conditional because Postgres treats NULLs as distinct, so one key including `opinion` would place no limit at all on volume rows
- a check constraint requiring an opinion for the opinion-level stages and forbidding one for the others. Without it, an extract row with no opinion attributes one opinion's work to the whole volume and collides with its siblings
- `OPINION_LEVEL_STAGES` is a tuple rather than a frozenset, because it is embedded in that constraint and an unordered container would rewrite the migration on every run

**`run` and `attempt` are different things.** `run` is a stage-run generation, scoped per (stage, engine, opinion): re-running one engine must not renumber the other, or a stage barrier reading "rows at max(run)" stops seeing the engine that was not re-run. `attempt` increments on every resubmission of the same row and is what the result key carries, so an abandoned worker's late upload cannot be harvested as the current attempt's output. `retry_count` stays pure accounting.

**`COMPLETED` is open, not terminal.** `OPEN_JOB_STATUSES` includes it. The provider being finished is not the same as us having applied the result, and treating the row as done at that point is exactly how a finished job's output gets lost.

**`input_key` plus `source_digest` are provenance, not a reference.** Opinion PDFs are renamed when pairing moves a boundary, so the FK alone cannot say which file version was read, or whether an extraction still describes the current one.

### Deliberately not here

- **No shard tables and no page ranges.** Extraction runs per opinion PDF on files that already exist, after every page fix, so there is nothing to split and no offset that can drift.
- **No bitonal.** It runs in-process via blackletter and fans out over local workers (#162), so it is not an external job. Only work that leaves the process gets a row.
- **No doctor provider** (#158), no bulk-batch table, and no inline result path. Results go to S3 only, via the presigned PUT contract from #157, so there is one harvest branch and one staleness rule.

Each of those is additive if it comes back.

### Two things the integration PR has to settle

- **Result key scoping.** #159 shipped `s3_job_result_key(scan, stage)` deliberately *not* run-scoped, and #161 builds `reusable_result` on that stable key. This model's contract needs keys scoped to run and attempt. The intended replacement is that a re-validate bumps `run` instead of calling `invalidate_job_results`, and resume becomes a row read rather than an S3 timestamp probe. #161 should land first and be simplified afterwards, not blocked.
- **#165 is a prerequisite for the extract stage.** `run_generate_files` deletes and recreates a scan's `OpinionScan` rows, so with the FK cascading, regenerating files discards every extraction job for the volume. CASCADE is still right (SET_NULL leaves rows violating the stage-shape invariant and miscounts the barrier; PROTECT breaks generate_files outright), so the fix belongs in generate_files. A test documents the current behaviour so it cannot surprise anyone later.

## Testing

40 new tests in `scanning/tests/test_external_jobs.py`, covering both stage shapes, every constraint, the status sets, deadlines, run numbering, and attempt history. Full suite: 499 passed, 1 skipped. `makemigrations --check` clean.

## AI Disclosure

- [ ] No AI tools were used to create the content of this PR.
- [x] Parts of this PR were created with the help of an AI tool, and I have carefully reviewed all of its content and take full responsibility for it.


scanning/models.py
```diff
@@ -1255,3 +1255,619 @@ class PendingUpload(AbstractDateTimeModel):
...
+    blackletter, which fans out over local workers by itself, so it is
+    not an external job and owns no row here.
+    """
+
+    BLACKLETTER = "blackletter", "blackletter (YOLO + PaddleOCR)"
+    DOTS_MOCR = "dots_mocr", "dots.mocr"
+    MISTRAL_OCR = "mistral_ocr", "Mistral OCR"
+    SURYA = "surya", "Surya"
+    LIGHTON_OCR = "lighton_ocr", "LightOnOCR"
+
```
grossir, line 1293: I think we should add bitonal. And update the comment

scanning/models.py
```diff
@@ -1255,3 +1255,619 @@ class PendingUpload(AbstractDateTimeModel):
...
+    RunPod hands back a job id and answers ``GET /status``, while a
+    hosted API answers on its own endpoint in its own shape. Keeping
+    this apart from the engine means the several models we run on RunPod
+    share one polling implementation, and moving one of them elsewhere
+    is a value change rather than a new branch at every call site.
+    """
+
+    RUNPOD = "runpod", "RunPod Serverless"
+    MISTRAL = "mistral", "Mistral API"
+
```
grossir, line 1276: let's add doctor for bitonal

scanning/models.py
```diff
@@ -1255,3 +1255,619 @@ class PendingUpload(AbstractDateTimeModel):
...
+    shard_index = models.PositiveIntegerField(
+        default=0,
+        help_text=(
+            "Position of this job within its target's fan-out, and part "
+            "of the unique key. Zero of one for the normal case, since "
+            "an opinion PDF is read whole; the split exists so a "
+            "document too large for one request can be divided without "
+            "a schema change."
+        ),
+    )
```
grossir, line 1557: I think this name is confusing, or at least the `help_text` is . I think it would help a lot to write it ourselves, Claude's writing is verbose and a little convoluted

What I am understanding is that this is covering cases like a volume that has been scanned into 2 PDFs. 
But one of our regular cases is **sharding for speed** (both bitonal and dots.mocr), not necessarily because the document is too big itself

scanning/models.py
```diff
@@ -1255,3 +1255,619 @@ class PendingUpload(AbstractDateTimeModel):
...
+        max_length=64,
+        blank=True,
+        default="",
+        help_text=(
+            "Digest of the document at ``input_key`` when it was sent. "
+            "Provenance rather than a reference, so it answers the "
+            "question the FK cannot: whether an extraction still "
+            "describes the current file, or the opinion has been "
+            "regenerated underneath it and has to be read again."
+        ),
```
grossir: this is the input hash, right? I think it would be better named like that `input_hash` matching `input_key`

scanning/models.py
```diff
@@ -1255,3 +1255,619 @@ class PendingUpload(AbstractDateTimeModel):
...
+    )
+    input_manifest = models.JSONField(
```
grossir, line 1585: can you give an example of how this would look?

grossir (commented):
This looks good. Just some naming / help_text observations @quevon24

scanning/models.py
quevon24: Added an example to the help_text

`{"crops": [{"key": "page_0007_84_132_1620_230", "page_index": 7, "bbox": [84, 132, 1620, 230]}]}`

scanning/models.py
quevon24: Renamed to `input_hash`

quevon24:
Thanks @grossir, all applied.

On the wording, you were right about more than the three fields you flagged, so I did a pass over the whole model

scanning/models.py
```diff
@@ -1255,3 +1255,611 @@ class PendingUpload(AbstractDateTimeModel):
...
+            "it, worker metadata, timings, sizes, cost, and an "
+            "append-only ``attempts`` list of prior attempts so "
+            "mutating this row on retry still keeps its history."
+        ),
+    )
+
+    objects = ExternalJobQuerySet.as_manager()
+
+    class Meta:
+        ordering = ["scan", "stage", "run", "engine", "opinion", "shard_index"]
```
grossir: this should order by `scan_id` and by `opinion_id` not by `scan` and `opinion`. Otherwise, it will produce joins each time ExternalJob is queried. You can inspect the SQL produced by the queryset

```sql
  LEFT OUTER JOIN "scanning_opinionscan"
          ON ("scanning_externaljob"."opinion_id" = "scanning_opinionscan"."id")
  ORDER BY "scanning_scan"."date_created" DESC,          -- "scan"    → Scan.Meta.ordering                                                                                   
           "scanning_externaljob"."stage" ASC,
           "scanning_externaljob"."run" ASC,
           "scanning_externaljob"."engine" ASC,
           "scanning_opinionscan"."opinion_order" ASC,   -- "opinion" → OpinionScan.Meta.ordering
           "scanning_opinionscan"."date_created" DESC,   --   (both of its ordering fields)
           "scanning_externaljob"."shard_index" ASC

```

vs

```sql
  FROM "scanning_externaljob"
  ORDER BY "scanning_externaljob"."scan_id" ASC,
           "scanning_externaljob"."stage" ASC,
           "scanning_externaljob"."run" ASC,
           "scanning_externaljob"."engine" ASC,
           "scanning_externaljob"."opinion_id" ASC,
           "scanning_externaljob"."shard_index" ASC

```
Also implicit on the [docs](https://docs.djangoproject.com/en/6.1/ref/models/options/#ordering) warning box

grossir (commented):
Looks good. Just one optimization comment @quevon24

scanning/models.py
quevon24: Good catch, fixed. Ordering is now by `scan_id`/`opinion_id`, and the queryset is a plain single-table select again. Same problem exists on `Page.Meta.ordering`.

grossir (approved):
LGTM

PR #73, opened 2026-05-20
feat(scanning): add confirmation dialog to RunPod buttons

quevon24:
Closes #60.

Adds a confirmation dialog to the four buttons in `/process/` that enqueue RunPod jobs: Validate, Re-validate, Next: Detect →, and Rebuild & Validate. A small cloud icon marks each button, and both the hover tooltip and `confirm()` message flag the cost ($$$ for definite RunPod usage).

The icon and confirm are skipped on no-cost paths:

- Rebuild & Validate when there are only pending page deletions (no inserts).
- Next: Detect → when Detection rows already exist (the view just redirects to step 2 without re-running detection).

PR #167, opened 2026-08-12
feat: swap detection to bl-warm, the single 18-class model

rachlllg:
## Fixes
Companion to freelawproject/blackletter#73 (the bl-warm adapter this depends on). No scanning issue on file.

## Summary
This PR swaps the detection step from the legacy small/medium/large ensemble to bl-warm, a single 18-class model, and ships a standalone viewer for judging the swap.

**Production diff (two settings + wiring, ~30 lines).** `YOLO_DETECT_MODELS` selects the blackletter weights the detection step runs (default `["bl_warm"]`), and `YOLO_DETECT_ON_BITONAL` selects its input (default off): the legacy trio detects on the bitonal processing copy, while bl-warm was trained on grayscale renders and must detect on the original upload — its large region classes collapse on 1-bit pages (measured caption F1 0.99 → 0.25). Bitonal remains in use for tesseract OCR and the output PDFs. The legacy configuration stays one env var away as the rollback path. Everything downstream — detections sidecar, redaction rects, headnote snap/split, margins, generate — is unchanged.

**Evidence.** End-to-end through this repo's redaction flow (replicated step-for-step in the evaluation harness), scored against 270 hand-annotated golden pages from four held-out volumes:

| macro F1 @IoU 0.5 | val (188 pages) | test (82 pages) |
|---|---|---|
| legacy 3-model ensemble (bitonal) | 0.727 | 0.705 |
| **bl-warm (original input)** | **0.965** | **0.907** |

Ten of sixteen scored classes reach F1 1.000 on test; the biggest gains are caption, keycite, case_sequence, page_number, syllabus, and headnote_bracket.

**Comparison viewer (`comparison/`).** A standalone FastAPI app (own uv project, not part of the Django deployment) with two views: per-page class detections (ground truth vs bl-warm vs the legacy ensemble, per-pane dropdowns) and full-volume redaction browsing (unredacted / bl-warm / legacy, three synced panes, ~2,700 pages per version). Its 2.9G data folder is gitignored and shared separately as a zip; unzip so `data/` sits next to `app.py` and run `uv run uvicorn app:app --port 8990`.

**Gotchas.**
- `pyproject.toml` pins `blackletter>=0.2.0`, which predates the bl_warm adapter; the swap needs a blackletter release containing it (freelawproject/blackletter#73), and the RunPod worker image must be rebuilt against that release. Until then the default settings would fail to resolve the `bl_warm` weight on a worker.
- With `YOLO_DETECT_ON_BITONAL=False`, workers fetch the original PDF rather than the ~20× smaller bitonal copy, so per-scan transfer grows accordingly.

Test plan:

- [x] Full test suite passes with the diff (the failing set is byte-identical with and without these changes — pre-existing env gaps: `requests` and the `local-ml` extra not installed locally).
- [x] E2E parity run over four full golden volumes with the exact production flow replicated (numbers above); redacted output inspected page-by-page in the comparison viewer.

## AI Disclosure
- [ ] No AI tools were used to create the content of this PR.
- [x] Parts of this PR were created with the help of an AI tool, and I have carefully reviewed all of its content and take full responsibility for it.


quevon24:
Rebuilt and released the RunPod worker image, so detection now actually runs bl-warm. The deployed image had `blackletter` 0.1.1 from PyPI, which has no `bl_warm` adapter or weight source, so every detect job failed. bf6509d fixes the worker pin and bakes `bl_warm.pt` in.

Note that pin is a git rev of freelawproject/blackletter#73, not a release. It has to go back to a PyPI pin (worker + root `pyproject.toml`, plus a relock) once #73 ships, before this can merge.

scanning/runpod/Dockerfile
```diff
@@ -104,7 +109,7 @@ RUN set -e; \
 import blackletter, pathlib; \
 from ultralytics import YOLO; \
 wd = pathlib.Path(blackletter.__file__).resolve().parent / 'weights'; \
-sizes = {'small.pt': 15_000_000, 'medium.pt': 40_000_000, 'large.pt': 100_000_000}; \
+sizes = {'small.pt': 15_000_000, 'medium.pt': 40_000_000, 'large.pt': 100_000_000, 'bl_warm.pt': 45_000_000}; \
```
grossir: no use on downloading small, medium, large

grossir (commented):
We will reuse parts of this PR once we are done with #190  I am not closing it until we begin implementation. Some notes below:

The comparison app should be a branch not part of the PR, since comparing the legacy and current approaches is not part of the main product.

We should split this into different PRs so it is tractable

- build the new runpod image
- integrate this RunPod call to the external job model
    - this step is mostly not handled by this PR
- actually use and serve the redaction results
    - this step will reuse most of this PR

PR #169, opened 2026-08-13
Shard the original PDF for external job processing

### LINKED ISSUE #164: Evaluate scan PDF sharding strategy for external job processing
We need to split the original scan PDF into smaller PDFs for external job execution. 

Currently, the ones we know benefit / need sharding are
- bitonal on doctor #158 
- dots.mocr on Runpod

Need to research the benefits, if any, of sharding
- MistralOCR API ([reported](https://flp-talk.slack.com/archives/C0B9XMW1X6J/p1782233200860729) to be very fast 0.3s/page) may not need sharding
- Surya (will it be run through an API?)
- YOLO

### Advantages

We expect that sharding and running the jobs in parallel will be:
- faster
- safer (smaller cost to retry)

About "faster"
- bitonal [takes](https://github.com/freelawproject/scanning/discussions/144#discussioncomment-17832530) 6.3 minutes on average for a full volume on our pod
- About dots.mocr: On #130 Kevin found 5.3 s / page  for a 30 page document. Rachel had 3/s a page. Note that hardware was different accross measurements. However, for a 1000 page volume, on both instances it's too much to wait.
- Additional time costs may come from uploading. However,  in #124 there is an estimated 385 MB/s for uploads from the server "Streamed 2004.5 MB … in 5.2s". So the extra upload time should be a few seconds at most
- cold start time will be tracked here https://github.com/freelawproject/scanning/issues/156 . More of a "how to use sharded PDFs" question than a "how to shard the PDF" question, but important to note.

### Open questions about sharding itself

Points to analyze before implementing:
- how to shard? Mike suggest using pdftk. Are there other options? Whatever tool we use, it should keep the exact characteristics of the input PDF 

- ideal shard size in bytes for each provider: bitonal and dots.mocr. Ideally, we can find a midpoint between the two. Byte size because some PDFs may hold images that make sizing different
        - In the case of bitonal, since we will run it on doctor #158 , we need to confirm the current deployment can support the shard size

- sharding costs in processing time

### Pipeline integration
- This should run as soon as the Scan is uploaded, since we need dots.mocr output as soon as we can for the first human review
- sanity checks to ensure sharding is equivalent to the original file
- what happens if sharding fails?
- once the shards are computed, we should never recompute them
- what prefix to use in S3, under the scan? 
- these files should not be auto synced when interacting with a scan, only when they are needed for processing
- we should add logging to evaluate shard computation time and upload time.

grossir:
Refs #164 (leaving it open for the consumer-side questions: per-provider shard sizing, additions-as-standalone-shards, output composition).

Implements the sharding mechanics so external workers (bitonal on doctor #158, dots.mocr on RunPod) can process page ranges of a volume in parallel instead of the whole multi-GB original.

## What it does

At the start of `run_full_pipeline` (queued on upload for the upload-and-validate flow), `sharding.ensure_shards(scan)` cuts the original into ~`SHARD_TARGET_BYTES` (200 MiB, ~127 pages at the corpus median) shards with PyMuPDF, verifies them, and uploads them under `shards/` in the scan's S3 processing prefix next to a `manifest.json`.

- **Split**: shard count from bytes (`ceil(size / target)`), pages spread evenly — no tiny tail shard. ~3 s per volume, byte-identical page copies (per the [#164 benchmark](https://github.com/freelawproject/scanning/issues/164#issuecomment-5259778865), 20/20 tool pairs round-tripped at +0.0%).
- **Verify**: per-page MediaBox equality + digests of the raw undecoded image streams — the one check that catches a tool/version silently recompressing scan data.
- **Manifest as commit marker**: uploaded last, so its presence guarantees every shard it lists is in the bucket. No manifest survives a failure, so a re-queued scan retries the sharding.
- **Sync exclusion**: `shards/` is excluded from the generic processing sync in both directions (like `jobs/`) — the shards duplicate the original's bytes. Local copies are dropped after upload; consumers fetch via presigned GET.

## Design decisions (from discussion on #164)

- **Shards are a fingerprint-keyed cache for full-volume jobs only.** The manifest records the source's size + page count. `ensure_shards` reuses the set while the fingerprint matches and sweeps + re-cuts when it doesn't (page edits accumulated, re-upload).
- **Smart page inserts/deletes never touch shards.** All page processing is page-level: an edited page is processed individually and merged into the volume-level artifacts, as the smart-edit paths already do. The stale set just sits in S3 until the next full-volume run (user-facing "Re-validate", admin re-queue) replaces it — a ~3 s cost buried under the full RunPod work the user explicitly confirmed. **Any future consumer must go through `ensure_shards` first — never read `shards/` directly.**
- **Sharding is a regular pipeline stage**: failures propagate to `_handle_pipeline_exception` and mark the scan ERROR. `SHARDING_ENABLED` (default on) is the rollout kill switch.
- **Retuning `SHARD_TARGET_BYTES` does NOT re-shard existing volumes** — only a source change does.

## Notes for review

- 200 MiB was sized against doctor's measured limits (memory is not the constraint if bitonal streams page-by-page; ~45 s/shard). dots.mocr at 3–5.3 s/page makes a ~127-page shard a 6–11 min job — may be worth revisiting the target when that lands; new volumes pick up a retuned value automatically.
- Shards persist in S3 indefinitely (~1× the original per volume, ≈$0.05/volume/month). Deleting after consumption and re-cutting on demand is a possible follow-up if the bucket bill ever matters.
- `upload_only` scans don't shard until someone queues processing — acceptable, since first-review urgency only applies to scans headed for review.
- If page additions are later processed as standalone mini-shards (per [the composition comment](https://github.com/freelawproject/scanning/issues/164#issuecomment-5275464822)), they must not live under `shards/` unless added to the manifest: the manifest-lists-everything invariant and the pre-reshard sweep both assume `ensure_shards` wholly owns that prefix.

## Test plan

- `scanning.tests.test_sharding` (19 tests): split math edge cases, round-trip fidelity, tamper/missing/gap detection, idempotent reuse, stale-fingerprint re-shard, failure cleanup, disabled mode, pipeline wiring (`run_full_pipeline` runs the stage; failures propagate).
- `scanning.tests.test_s3_sync` (+7): sync exclusions in both directions, manifest-last upload ordering, prefix sweep, manifest fetch/parse.
- Full suite: 485 tests, OK.


grossir:
@claude review

grossir:
First statistics coming in

Processing scan 1948 (full_pipeline)
INFO Downloaded 1 processing file(s) for scan 1948 from s3://com-courtlistener-scanning-private-storage/processing/1948/p3d/470/1/
INFO Sharding...
INFO Uploaded 14 shard file(s) (1953.2 MB) for scan 1948 to s3://com-courtlistener-scanning-private-storage/processing/1948/p3d/470/1/shards/ in 30.6s (64 MB/s)
INFO Sharded scan 1948: 13 shard(s) of ~100 page(s) from 1953.1 MB / 1291 pages, split 11.8s, verify 82.8s, uploaded to S3
INFO Sharding done (125.9s)

